PR #6, opened 2016-07-12
Upgrade to OpenTracing 1.1 with SpanContext

yurishkuro:


jaeger_client/utils.py
```diff
@@ -61,3 +63,33 @@ def get_boolean(string, default):
         return True
     else:
         return default
+
+
+def local_ip():
+    """Get the local network IP of this machine"""
+    ip = socket.gethostbyname(socket.gethostname())
```
thepaulm: This code causes pain right now in its current instantiation in tchannel. Lot's of laptops fail calls to socket.gethostname() because of ad-hoc ephemeral nature of the their hostname from the wifi. It's become common practice for devs to hard code their hostname and then add it to their /etc/hosts file to make this work.

I'd rather not have people who are approaching this code for the first time run into this. Can we perhaps do something like this:

`
try:
    ip = socket.gethostbyname(socket.gethostname())
except:
    ip = socket.gethostbyname('localhost')

`

thepaulm:
Other than my request about the local_ip() function, this looks great!

PR #13, opened 2016-08-23
Implement tchannel crossdock tests

black-adder:


crossdock/server/serializer.py
```diff
@@ -30,11 +30,13 @@ def downstream_from_struct(i):
     return d
 
 
-def downstream_to_json(d):
-    s = {}
+def join_trace_request_to_json(d, s):
```
yurishkuro: please use readable variable names

crossdock/server/server.py
```diff
@@ -88,14 +89,42 @@ def tchannelNotImplResponse():
...
+def make_tchannel(port):
+    global tchannel
+    tchannel = TChannel('python', hostport="localhost:%d" % port, trace=False)
+
+    service = get_thrift_service(service_name='python')
+
+    @tchannel.thrift.register(service.TracedService, method='startTrace')
+    @tornado.gen.coroutine
+    def start_trace(request):
+        return trace_response_to_thriftrw(service, tchannelNotImplResponse())
```
yurishkuro: may want to raise exception instead, the `tchannelNotImplResponse()` is not the right return type anyway for `start_trace`

crossdock/server/server.py
```diff
@@ -118,91 +147,104 @@ def update_span(span):
...
-    def call_downstream_tchannel(self, span, trace_request, trace_response, response_writer):
-
-        tr = tchannelNotImplResponse()
-        response_writer.write(serializer.obj_to_json(tr))
+        raise UnknownTransportException("%s" % trace_request.downstream.transport)
+    raise tornado.gen.Return(resp)
+
+
+@tornado.gen.coroutine
+def call_downstream_http(span, trace_request, trace_response, response_writer):
```
yurishkuro: are you re-implementing that method?  It was already working before.

black-adder: I expanded the function to receive and send thriftrw over http

crossdock/server/server.py
```diff
@@ -88,14 +95,46 @@ def tchannelNotImplResponse():
...
 
+    service = get_thrift_service(service_name='python')
 
-@tchannel.thrift.register(service.TracedService)
-def joinTrace(request):
-    return trace_response_to_thriftrw(service, tchannelNotImplResponse())
+    @tchannel.thrift.register(service.TracedService, method='startTrace')
+    @tornado.gen.coroutine
+    def start_trace(request):
+        raise UnimplementedEndpointException("tchannel startTrace endpoint not implemented")
```
yurishkuro: just don't add this endpoint to begin with, tchannel will throw some exception automatically. And we won't need UnimplementedEndpointException

crossdock/server/server.py
```diff
@@ -130,79 +168,88 @@ def handle_trace_request(self, http_request, trace_request, span_handler, respon
         if span_handler:
             span_handler(span)
 
-        trace_id = "%x" % span.trace_id
-        observed_span = ObservedSpan(
-            trace_id, span.is_sampled(),
-            span.get_baggage_item(constants.baggage_key))
-
-        tr = TraceResponse(span=observed_span)
+        with tchannel.context_provider.span_in_context(span):
```
yurishkuro: while functionally the same, this is an HTTP server and it's better if it used `span_in_stack_context` method from `opentracing_instrumentation` (more testing this way)

crossdock/server/server.py
```diff
@@ -105,8 +144,7 @@ class UnknownTransportException(Exception):
 
 class Server(object):
     def __init__(self):
-        self.tracer = Tracer("python", None, ConstSampler(decision=True))
-        opentracing.tracer = self.tracer
+        self.tracer = opentracing.tracer
 
     def start_trace(self, request, response_writer):
         sstr = serializer.start_trace_request_from_json(request.body)
```
yurishkuro: not sure what sstr stands for. Can we use `start_trace_req`?

crossdock/server/server.py
```diff
@@ -130,79 +168,88 @@ def handle_trace_request(self, http_request, trace_request, span_handler, respon
         if span_handler:
```
yurishkuro: I suggest accepting `downstream`, not `trace_request`, because the latter is being overloaded by both start and join requests that happened to share `downstream` property. But in practice we don't care about top level request here, only the downstream.

crossdock/server/server.py
```diff
@@ -130,79 +168,88 @@ def handle_trace_request(self, http_request, trace_request, span_handler, respon
...
-        client = tornado.httpclient.AsyncHTTPClient()
-        client.fetch(req, handle_response)
+                response_writer.finish()
 
-    def call_downstream_tchannel(self, span, trace_request, trace_response, response_writer):
 
-        tr = tchannelNotImplResponse()
-        response_writer.write(serializer.obj_to_json(tr))
+@tornado.gen.coroutine
+def call_downstream(span, trace_request, trace_response, response_writer):
```
yurishkuro: 1. similar to above - we don't need `trace_request` here, only `downstream`
 1. let's have the functions without side effects - the should return downstream response, not modify the input parameter.

crossdock/server/server.py
```diff
@@ -130,79 +168,88 @@ def handle_trace_request(self, http_request, trace_request, span_handler, respon
...
 
-    def call_downstream(self, span, trace_request, trace_response, response_writer):
-        if trace_request.downstream.transport == Transport.HTTP:
-            self.call_downstream_http(span, trace_request, trace_response, response_writer)
-        elif trace_request.downstream.transport == Transport.TCHANNEL:
-            if tchannel_supported:
-                self.real_call_downstream_tchannel(span, trace_request, trace_response,
-                                                   response_writer)
+            if trace_request and trace_request.downstream is not None:
+                call_downstream(span, trace_request, tr, response_writer)
```
yurishkuro: call_downstream is a coroutine, how are we calling it here as a plain function? And shouldn't the endpoint handling methods also be coroutines?

crossdock/server/server.py
yurishkuro: I would rather separate responsibilities. The endpoint should be responsible for generating response data and writing it out to `response_writer`. There's no reason why `response_writer` should be exposed to anyone else, as it leads to this ugly branching in this function as it tries to do things it should not be responsible for. All it needs to do is
 * format the request to json
 * call downstream
 * format response back into object

setup.py
```diff
@@ -41,6 +41,8 @@
         'thrift',
         'tornado>=4.3,<5',
         'opentracing>=1.1,<1.2',
+        'tchannel>=0.27',
+        'opentracing_instrumentation>=2,<3',
```
yurishkuro: we do not want jaeger client itself to depend on these libs

yurishkuro:
Q: somewhat suspicious about coverage numbers not changing at all. Are the crossdock files being tested?

crossdock/server/server.py
```diff
@@ -77,25 +84,55 @@ def head(self):
...
 
-@tchannel.thrift.register(service.TracedService)
-def startTrace(request):
-    return trace_response_to_thriftrw(service, tchannelNotImplResponse())
+    service = get_thrift_service(service_name='python')
 
+    @tchannel.thrift.register(service.TracedService, method='joinTrace')
+    @tornado.gen.coroutine
+    def join_trace(request):
+        jtr = request.body.request if request.body.request else None
```
yurishkuro: nit: `request.body.request or None`

crossdock/server/server.py
```diff
@@ -77,25 +84,55 @@ def head(self):
...
+
+        if trace_request and trace_request.downstream is not None:
+            downstream_trace_resp = yield call_downstream(span, trace_request.downstream)
+            observed_span = service.ObservedSpan(
+                traceId=downstream_trace_resp.span.traceId,
+                sampled=downstream_trace_resp.span.sampled,
+                baggage=downstream_trace_resp.span.baggage
+            )
+            downstream_trace_resp = TraceResponse(span=observed_span, notImplementedError='')
+            downstream_trace_resp = trace_response_to_thriftrw(service, downstream_trace_resp)
```
yurishkuro: this function again looks like it modifies its arguments, can it not simply return TraceResponse?

crossdock/server/server.py
```diff
@@ -130,79 +166,66 @@ def handle_trace_request(self, http_request, trace_request, span_handler, respon
...
 
-        trace_id = "%x" % span.trace_id
-        observed_span = ObservedSpan(
-            trace_id, span.is_sampled(),
-            span.get_baggage_item(constants.baggage_key))
+        with request_context.span_in_stack_context(span):
+            trace_id = '%x' % span.trace_id
+            observed_span = ObservedSpan(
+                trace_id, span.is_sampled(),
+                span.get_baggage_item(constants.baggage_key))
```
yurishkuro: this can be DRY-ed up with a helper function

crossdock/server/server.py
```diff
@@ -130,79 +166,66 @@ def handle_trace_request(self, http_request, trace_request, span_handler, respon
...
-        else:
-            response_writer.finish()
-            raise UnknownTransportException("%s" % trace_request.downstream.transport)
+            if trace_request and trace_request.downstream is not None:
 
-    def call_downstream_http(self, span, trace_request, trace_response, response_writer):
+                def handle_response(future):
+                    downstream_trace_resp = future.result()
+                    if downstream_trace_resp.notImplementedError:
+                        response_writer.write(serializer.obj_to_json(downstream_trace_resp))
```
yurishkuro: this branch is not needed, we simply pass downstream result as part of our own result. There used to be a requirement to propagate "not-implemented" explicitly, but now go-client takes care of that

crossdock/server/server.py
```diff
@@ -130,79 +166,66 @@ def handle_trace_request(self, http_request, trace_request, span_handler, respon
...
+                        trace_response.downstream = downstream_trace_resp
+                        response_writer.write(serializer.obj_to_json(trace_response))
+                    response_writer.finish()
 
-        def handle_response(response):
-            tr = serializer.traceresponse_from_json(response.body)
-            if tr.notImplementedError:
-                response_writer.write(serializer.obj_to_json(tr))
+                future = call_downstream(span, trace_request.downstream)
+                tornado.ioloop.IOLoop.current().add_future(future, handle_response)
```
yurishkuro: this is not necessary, let's just have all methods be normal coroutines and you can simply call `res = yield func()` instead of coding with callbacks (which are always harder to read)

crossdock/server/server.py
```diff
@@ -130,79 +166,66 @@ def handle_trace_request(self, http_request, trace_request, span_handler, respon
...
+
+@tornado.gen.coroutine
+def call_downstream_http(span, downstream):
+    url = 'http://%s:%s/join_trace' % (downstream.host, downstream.port)
+    body = serializer.join_trace_request_to_json(downstream.downstream, downstream.serverRole)
+
+    req = tornado.httpclient.HTTPRequest(url=url, method='POST',
+                                         headers={'Content-Type': 'application/json'},
+                                         body=body)
+    http_client.before_http_request(tornado_http.TornadoRequestWrapper(request=req),
```
yurishkuro: doesn't this return a span that you need to finish after the call?

crossdock/server/server.py
```diff
@@ -130,79 +166,66 @@ def handle_trace_request(self, http_request, trace_request, span_handler, respon
...
+                                    lambda: span)
+    client = tornado.httpclient.AsyncHTTPClient()
+    future = yield client.fetch(req)
```
yurishkuro: something odd here, yield should not be returning future, but the actual data. Perhaps variable name is wrong, since you call `.body` on it - it's probably just an http response.

crossdock/server/server.py
```diff
@@ -130,79 +166,66 @@ def handle_trace_request(self, http_request, trace_request, span_handler, respon
...
+        downstream_trace_resp = yield call_downstream_tchannel(downstream)
+    else:
+        raise UnknownTransportException('%s' % downstream.transport)
+    raise tornado.gen.Return(downstream_trace_resp)
+
+
+@tornado.gen.coroutine
+def call_downstream_http(span, downstream):
+    url = 'http://%s:%s/join_trace' % (downstream.host, downstream.port)
+    body = serializer.join_trace_request_to_json(downstream.downstream, downstream.serverRole)
```
yurishkuro: args order not consistent, lets (a) keep the order matching thrift class definition (I think server role is first), and (b) use name=value parameter passing. Also, let's have parity with `join_trace_request_to_thriftrw` - that one takes the whole request, not two params, we should be consistent.

yurishkuro: actually I see why the other one does it, because of thrift - fine.

crossdock/server/server.py
```diff
@@ -130,79 +166,66 @@ def handle_trace_request(self, http_request, trace_request, span_handler, respon
...
+
+
+@tornado.gen.coroutine
+def call_downstream_tchannel(downstream):
+    downstream_service = get_thrift_service(downstream.serviceName)
+
+    jtr = JoinTraceRequest(downstream.serverRole, downstream.downstream)
+    jtr = join_trace_request_to_thriftrw(downstream_service, jtr)
+
+    future = yield tchannel.thrift(downstream_service.TracedService.joinTrace(jtr),
```
yurishkuro: future or thriftResult?

crossdock/server/server.py
black-adder: I attempted this and got: "StackContextInconsistentError: stack_context inconsistency (probably caused by yield within a "with StackContext" block)"
Is there a way around this without having to pass down span?

yurishkuro: it should be documented in the method, the pattern is
```python
with span_in_stack_context(span=span):
   future = call_coroutine()
result = yield future
```

i.e. you do the yield outside of the `with:` statement

crossdock/server/serializer.py
```diff
@@ -76,15 +78,15 @@ def class_keys(obj):
 def obj_to_json(obj):
```
yurishkuro: which `obj` is this?

crossdock/server/server.py
```diff
@@ -77,25 +83,41 @@ def head(self):
...
+            (r'/start_trace', MainHandler, (dict(server=server, method=Server.start_trace))),
+            (r'/join_trace', MainHandler, (dict(server=server, method=Server.join_trace))),
         ], debug=True)
 
 
-# TChannel handlers
-def tchannelNotImplResponse():
-    return TraceResponse(notImplementedError="python <=> tchannel not implemented")
+def make_tchannel(port):
+    global tchannel
```
yurishkuro: you're returning `tchannel` from the function, does it have to be global?

crossdock/server/server.py
```diff
@@ -77,25 +83,41 @@ def head(self):
...
+
+    @tchannel.thrift.register(service.TracedService, method='joinTrace')
+    @tornado.gen.coroutine
+    def join_trace(request):
+        join_trace_request = request.body.request or None
+        response = yield handle_downstream_request(join_trace_request.downstream)
+        raise tornado.gen.Return(response)
+
+    @tornado.gen.coroutine
+    def handle_downstream_request(downstream):
```
yurishkuro: nit: this would be better named `prepare_response`, because it does not always do downstream

crossdock/server/server.py
```diff
@@ -77,25 +83,41 @@ def head(self):
...
+        trace_response = TraceResponse(span=observed_span, notImplementedError='')
 
-@tchannel.thrift.register(service.TracedService)
-def startTrace(request):
-    return trace_response_to_thriftrw(service, tchannelNotImplResponse())
+        if downstream:
+            downstream_trace_resp = yield call_downstream(span, downstream)
+            observed_span = observed_span_to_thriftrw(service, downstream_trace_resp.span)
+            downstream_trace_resp = TraceResponse(span=observed_span, notImplementedError='')
+            trace_response.downstream = trace_response_to_thriftrw(service, downstream_trace_resp)
```
yurishkuro: this section should only have two statements
```
downstream_response = yield call_downstream(span, downstream)
trace_response.downstream = trace_response_to_thriftrw(service, downstream_response)
```

crossdock/server/server.py
```diff
@@ -105,104 +127,93 @@ class UnknownTransportException(Exception):
...
-
-        self.handle_trace_request(request, jtr, None, response_writer)
+        response = yield self.handle_downstream_request(request, join_trace_request.downstream,
+                                                        None)
+        response_writer.write(response)
+        response_writer.finish()
 
-    def handle_trace_request(self, http_request, trace_request, span_handler, response_writer):
+    @tornado.gen.coroutine
+    def handle_downstream_request(self, http_request, downstream, span_handler):
```
yurishkuro: nit: `prepare_response`

yurishkuro: nit: s/span_handler/update_span_func/

every time I read this code, I stumble on it and have to look back what it means. Giving it a clear meaningful name makes the code much easier to read.

crossdock/server/server.py
yurishkuro: ping

crossdock/server/server.py
```diff
@@ -105,104 +127,93 @@ class UnknownTransportException(Exception):
...
 
-        jtr = JoinTraceRequest(trace_request.serverRole, downstream.downstream)
-        jtr = join_trace_request_to_thriftrw(service, jtr)
-        # with context.RequestContext(span):
-        with context.request_context(span):
-            f = tchannel.thrift(service.TracedService.joinTrace(jtr),
-                                hostport="%s:%s" % (downstream.host, downstream.port))
-        tornado.ioloop.IOLoop.current().add_future(f, handle_response)
+    thrift_result = yield tchannel.thrift(downstream_service.TracedService.joinTrace(jtr),
+                                          hostport='localhost:%s' % downstream.port)
```
yurishkuro: not localhost, `downstream.host`

crossdock/server/server.py
black-adder: the downstream response might be in json or thrift so I have to turn it into thrift here

crossdock/server/server.py
black-adder: yeah, the way it's setup right now means that the call_downstream_tchannel function needs the global tchannel function. I'll look into moving the tchannel setup and all the functions into the Server class

crossdock/server/server.py
black-adder: It does correctly send down the span to the downstream service.

crossdock/server/server.py
yurishkuro: this method is only called from tchannel join_trace handler, so I think there's only one form the request would come in. My preference is to have only a single method `prepare_response`, instead of two as you have, and have that method always deal with the same type in and out, most likely json classes. Then if that method is called from JSON handlers, you don't need to convert anything, otherwise the handler first converts thrift->json, calls prepare_response(), then converts result to thrift. Clean separation, DRY, no confusion which type is in or out.

You can see this in tchannel-python/crossdock server.

black-adder: I decided to pass everything around as a thrift objects around apart from in the handlers where I convert to json or thriftrw

crossdock/docker-compose.yml
```diff
@@ -22,7 +22,8 @@ services:
 
             - REPORT=compact
     go:
-        image: jaegertracing/xdock-go
+        # TODO Use the master branch once this branch is merged
+        image: jaegertracing/xdock-go:upgrade-crossdock-tests-for-native-opentracing-in-tchannel
```
yurishkuro: the Go repo has been upgraded, you can use the latest image now (rm the tag ^^^)

yurishkuro:
LGTM

Pull the latest Go img, and :shipit:

PR #23, opened 2016-10-28
Update idl

black-adder:


black-adder:
The thrift generated python code is pretty fragile and requires that the thrift is defined in a certain order. Once [jaeger-idl](https://github.com/uber/jaeger-idl/pull/14) lands, I'll update this diff

PR #25, opened 2016-10-31
Add Adaptive Sampler

black-adder:
- [x] Tests

This PR doesn't include initializing and updating the adaptive sampler. I'll do that in a subsequent PR.

### DIFF
--- a/jaeger_client/sampler.py
+++ b/jaeger_client/sampler.py
@@ -31,10 +31,12 @@ from .constants import (
     SAMPLER_TYPE_CONST,
     SAMPLER_TYPE_PROBABILISTIC,
     SAMPLER_TYPE_RATE_LIMITING,
+    SAMPLER_TYPE_LOWER_BOUND,
 )
 from .metrics import Metrics
 from .utils import ErrorReporter
 from .local_agent_net import parse_sampling_strategy
+from .thrift_gen.sampling.ttypes import PerOperationSamplingStrategies
 
 default_logger = logging.getLogger('jaeger_tracing')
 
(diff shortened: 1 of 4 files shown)


jaeger_client/sampler.py
```diff
@@ -177,6 +178,100 @@ def __str__(self):
...
+    """
+    def __init__(self, operation, lower_bound, rate):
+        super(GuaranteedThroughputProbabilisticSampler, self).__init__(
+            tags={
+                SAMPLER_TYPE_TAG_KEY: SAMPLER_TYPE_LOWER_BOUND,
+                SAMPLER_PARAM_TAG_KEY: rate,
+            }
+        )
+        self.probabilistic_sampler = ProbabilisticSampler(rate)
+        self.lower_bound_sampler = RateLimitingSampler(lower_bound)
```
yurishkuro: please double-check that this class works with values `<1`, as at some point the very first implementation (not sure in which language) only worked with qps expressed as a whole number

black-adder: I checked the code and it looks like it works. I added a test to make sure.

tests/test_sampler.py
```diff
@@ -116,13 +123,72 @@ def test_rate_limiting_sampler():
...
+
+def test_adaptive_sampler():
+    sampling_rates = [
+        OperationSamplingStrategy('op', ProbabilisticSamplingStrategy(0.5))
+    ]
+    strategies = PerOperationSamplingStrategies(0.51, 3, sampling_rates)
+
+    sampler = AdaptiveSampler(strategies, 2)
+    id1 = 1L << 63  # second most significant bit, dividing full range in half
+    assert id1 == 0x8000000000000000L
```
yurishkuro: we keep doing this, let's define  constant once and have a standalone test that validates that the constant is equal to `1L << 63` expression

tests/test_sampler.py
```diff
@@ -116,13 +123,72 @@ def test_rate_limiting_sampler():
...
+    # This operation is seen for the first time by the sampler
+    sampled, tags = sampler.is_sampled(id1-10, "new_op")
+    assert sampled
+    assert tags == get_tags('probabilistic', 0.51)
+
+    # This operation is seen for the first time by the sampler but surpasses
+    # max_operations of 2
+    sampled, tags = sampler.is_sampled(id1-10, "new_op_2")
+    assert sampled
+    assert tags == get_tags('probabilistic', 0.51)
```
yurishkuro: so there's no difference with new_op. Is there a way we can distinguish these two scenarios by asserting something else?

PR #26, opened 2016-11-03
Initialize adaptive sampler

black-adder:
- [x] Cover submit_callback

### DIFF
--- a/jaeger_client/sampler.py
+++ b/jaeger_client/sampler.py
@@ -394,11 +394,11 @@ class RemoteControlledSampler(Sampler):
 
     def _delayed_polling(self):
         periodic = PeriodicCallback(
-            callback=self.poll_sampling_manager,
+            callback=self._poll_sampling_manager,
             # convert interval to milliseconds
             callback_time=self.sampling_refresh_interval * 1000,
             io_loop=self.io_loop)
-        self.poll_sampling_manager()  # Initial sample now
+        self._poll_sampling_manager()  # Initial sample now
         with self.lock:
             if self.running:
                 self.periodic = periodic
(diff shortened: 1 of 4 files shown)


jaeger_client/sampler.py
```diff
@@ -340,11 +394,11 @@ def _init_polling(self):
 
     def _delayed_polling(self):
         periodic = PeriodicCallback(
-            callback=self.poll_sampling_manager,
+            callback=self._poll_sampling_manager,
             # convert interval to milliseconds
             callback_time=self.sampling_refresh_interval * 1000,
             io_loop=self.io_loop)
-        self.poll_sampling_manager()  # Initial sample now
+        self._poll_sampling_manager()  # Initial sample now
```
black-adder: I'm not sure what this comment is supposed to mean

jaeger_client/sampler.py
```diff
@@ -262,6 +290,29 @@ def is_sampled(self, trace_id, operation=''):
             return sampler.is_sampled(trace_id, operation)
         return sampler.is_sampled(trace_id, operation)
 
+    def update(self, strategies):
+        # (NB) This function should only be called while holding a Write lock.
```
badiib: if you have to lock, why not pass the lock in?

jaeger_client/sampler.py
```diff
@@ -353,40 +407,67 @@ def _delayed_polling(self):
...
-        fut.add_done_callback(submit_callback)
+        fut.add_done_callback(self._sampling_request_callback)
 
     def close(self):
         with self.lock:
             self.running = False
             if self.periodic is not None:
                 self.periodic.stop()
+
+    def _parse_sampling_strategy(self, sampler, body):
```
black-adder: I added the logic for operation_strategies

jaeger_client/sampler.py
yurishkuro: not sure, perhaps it meant "initialize sampler now", as it kicks off the very first refresh of the strategy and then starts periodic polling

jaeger_client/sampler.py
black-adder: look @ line 430, it's kinda weird to pass the lock to and fro. For new logic will be:

`if instance of AdaptiveSampler, then pass the lock inside; else lock and update sampler.`

Rather than just lock in one place.

jaeger_client/sampler.py
yurishkuro: can you fix the comment?

PR #30, opened 2017-02-15
Client enhancements

philipgian:
This PR add three things:
* The ability to override the codecs used by the tracer on initialization.
* The ability to override the default one-span-per-rpc behaviour on server-side spans.
  This is quite useful for visualizing the server-side spans on a call as nested spans.
* An interface and a mechanism that allows an external entity to accumulate and batch reporting spans for performance reasons.

### DIFF
--- a/jaeger_client/span.py
+++ b/jaeger_client/span.py
@@ -68,7 +68,7 @@ class Span(opentracing.Span):
             self.operation_name = operation_name
         return self
 
-    def finish(self, finish_time=None):
+    def finish(self, finish_time=None, defer_report=False):
         """Indicate that the work represented by this span has been completed
         or terminated, and is ready to be sent to the Reporter.
 
@@ -82,7 +82,8 @@ class Span(opentracing.Span):
             return
 
         self.end_time = finish_time or time.time()  # no locking
-        self.tracer.report_span(self)
+        if not defer_report:
+            self.tracer.report_span(self)
 
     def set_tag(self, key, value):
         """
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -46,7 +46,7 @@ class Tracer(opentracing.Tracer):
                  trace_id_header=constants.TRACE_ID_HEADER,
                  baggage_header_prefix=constants.BAGGAGE_HEADER_PREFIX,
                  debug_id_header=constants.DEBUG_ID_HEADER_KEY,
-                 one_span_per_rpc=True):
+                 one_span_per_rpc=True, extra_codecs=None):
         self.service_name = service_name
         self.reporter = reporter
         self.sampler = sampler
@@ -70,6 +70,8 @@ class Tracer(opentracing.Tracer):
             Format.BINARY: BinaryCodec(),
             ZipkinSpanFormat: ZipkinCodec(),
         }
+        if extra_codecs:
+            self.codecs.update(extra_codecs)
         self.tags = {
             constants.JAEGER_VERSION_TAG_KEY: constants.JAEGER_CLIENT_VERSION,
         }
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -45,7 +45,8 @@ class Tracer(opentracing.Tracer):
     def __init__(self, service_name, reporter, sampler, metrics=None,
                  trace_id_header=constants.TRACE_ID_HEADER,
                  baggage_header_prefix=constants.BAGGAGE_HEADER_PREFIX,
-                 debug_id_header=constants.DEBUG_ID_HEADER_KEY):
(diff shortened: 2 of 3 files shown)


jaeger_client/tracer.py
```diff
@@ -212,5 +216,8 @@ def _emit_span_metrics(self, span, join=False):
     def report_span(self, span):
         self.reporter.report_span(span)
 
+    def report_spans(self, spans):
```
yurishkuro: nothing calls this method, why do we need it?

jaeger_client/span.py
```diff
@@ -82,7 +82,8 @@ def finish(self, finish_time=None):
             return
 
         self.end_time = finish_time or time.time()  # no locking
-        self.tracer.report_span(self)
+        if not defer_report:
```
yurishkuro: why is this needed? What else would trigger reporting if not `span.finish()`?

yurishkuro:
Regarding the objectives of this PR:
  * the first two points would be best done as separate PRs, as otherwise it's hard to provide a sensible commit message describing the change
  * I don't quite follow why you need the last point. "an external entity to accumulate and batch reporting" is precisely what the `Reporter` interface is for. Specifically, the `RemoteReporter` does the batching and uses Senders to actually ship the data out of process. There may be a way to refactor the remote reporter to not depend on Tornado, with a separate thread - it was just easier to write with Tornado's producer-consumer framework.

yurishkuro:
SpanManager would be something that only knows about OpenTracing API. The Reporter and the relevant methods are the internal implementation detail of Jaeger tracer, SpanManager cannot make use of them. The only thing SpanManager can do is call `finish()` on the spans. So I do not thing any extension to reporting is needed.

philipgian:
I understand what you are saying, but I was aiming to have a SpanManager that checked if the tracer supports batch reporting and gave a hint to the Tracer that the Span will be reported later in a batch, otherwise fallback to standard OpenTracing API and just call `finish()`. So the SpanManager could work with all OpenTracing compatible Tracers, but work better with those supporting this "extension".

Anyway, if you object on having this merged, I will withdraw the changes and make only two new PRs with the first two bullets, and maybe proceed with a different approach.

yurishkuro:
I am definitely not in favor of extending the API, mostly because I still do not understand your use case. Maybe you have some instrumentation code you can demonstrate?  Whatever the responsibilities of your SpanManager are, it cannot control when `span.finish()` is called. For example, if a user wants to wrap a function into a span, they can do this:

```python
with tracer.start_span('my function', child_of=get_current_span()) as span:
    # do the work
    # maybe log to span
    span.log(..)
```

This will automatically call `finish()` when the span is logically completed, and neither the tracer nor the SpanManager can do anything to delay that. But internally the call to `finish()` simply delegates to `tracer.reportSpan()`, which passes the span to the reporter. If any batching needs to happen (and it already does in our RemoteReporter) it can happen inside the reporter.

philipgian:
SpanManager has to manage/accumulate/batch report only the spans it manages. And for those, it can require that they are finished only through it, and disallow manual `finish()` (e.g., by overriding the `finish()` function on the `Span` it returns). In your example, the span is created manually by the user and the SpanManager knows nothing about it, so of course it cannot control it.

The point is not to prevent or make batch reporting mandatory, but to facilitate towards this direction. I understand how the `Reporter()` solution works better, but I also consider the batch reporting support complementary, as it can also work with non batch-aware Reporters which are not  tied to a specific framework and which will be called only once in the lifetime of a request.

This extension was not meant to be compatible with a certain API, but enable certain functionality without breaking any API compatibility. Anyway, as I said, I totally understand your objection and I if your are not still convinced of the functionality it enables, I will withdraw this PR and work towards a different direction.

yurishkuro:
At this point I would advise perhaps just monkey-patching the methods onto the tracer and reporter in your own module. It goes without saying that you're making your module dependent on implementation details of Jaeger, as other OpenTracing tracers may not even have the concept of a "reporter" (at minimum it might be called differently). 

We try to keep to the "three strikes" rule, i.e. not introduce API changes or features unless there are 3 use cases for them. I am happy to revisit if you release your span manager code in the future so that we can talk in more concrete terms. 

In fact, if you think you're building some useful generic functionality, it might be worth opening an issue in opentracing-python repo (or even the specification repo) so that other people can weigh in. I think your use case sits between two audiences, the end users using OT for instrumentation on one hand, and the maintainers of the tracers on the other hand. So far OT has had no opinion on how the maintainers implement their OT API, this common functionality that affects how tracers work largely does not exist. For example, the process of registering custom codecs is implementation-specific.

philipgian:
OK. Thx for the comments. I'm closing this now as discussed. I've opened #32 as 1/2 PRs

PR #32, opened 2017-02-20
Optionally override one-span-per-rpc

philipgian:
Add option to override default one-span-per-rpc behaviour. This is
useful to visualize RPC calls as nested spans.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ History
 3.3.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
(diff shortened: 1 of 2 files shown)


yurishkuro:
Please merge or rebase from master to fix the build (cf. #33).

Also, you will need to sign a contributor agreement (once).

philipgian:
I've signed the CLA, but the relevant check is not updated. Does it require manual approval or is something wrong ?

yurishkuro:
It's possible the check only runs on push. Can you add a line to the CHANGELOG file, e.g. "Allow overriding one-span-per-rpc behavior" and push again?

PR #34, opened 2017-02-22
Add option to provide extra codecs

philipgian:
Add options to provide extra codecs to the tracer. This allows extending
or overriding the tracer's codec during initialization.


yurishkuro:
Can you add a unit test for the new option?

yurishkuro:
Awesome, thanks.

PR #36, opened 2017-03-15
Refactor remote sampler update logic

black-adder:
Existing logic could run into a NPE when updating the remote sampler from adaptive sampler to any other type. Refactored the logic to make it easier to reason about how the update is happening (just copied the Java implementation).

### DIFF
--- a/jaeger_client/sampler.py
+++ b/jaeger_client/sampler.py
@@ -153,6 +153,9 @@ class RateLimitingSampler(Sampler):
                 SAMPLER_PARAM_TAG_KEY: max_traces_per_second,
             }
         )
+        self._init_rate_limiter(max_traces_per_second)
+
+    def _init_rate_limiter(self, max_traces_per_second):
         assert max_traces_per_second >= 0, \
             'max_traces_per_second must not be negative'
         self.credits_per_second = max_traces_per_second
@@ -191,6 +194,13 @@ class RateLimitingSampler(Sampler):
         d1['balance'] = d2['balance']
         d1['last_tick'] = d2['last_tick']
         return d1 == d2
+
+    def update(self, max_traces_per_second):
+        # (NB) This function should only be called while holding a Write lock.
+        if self.credits_per_second == max_traces_per_second:
+            return
+        # only create a new rate limiter when the credits_per_second is changed
+        self._init_rate_limiter(max_traces_per_second)
 
     def __str__(self):
         return 'RateLimitingSampler(%s)' % self.credits_per_second
@@ -243,9 +253,8 @@ class GuaranteedThroughputProbabilisticSampler(Sampler):
                 SAMPLER_TYPE_TAG_KEY: SAMPLER_TYPE_LOWER_BOUND,
                 SAMPLER_PARAM_TAG_KEY: rate,
             }
-        if self.lower_bound != lower_bound:
-            self.lower_bound_sampler = RateLimitingSampler(lower_bound)
-            self.lower_bound = lower_bound
+        self.lower_bound_sampler.update(lower_bound)
+        self.lower_bound = lower_bound
 
     def __str__(self):
         return 'GuaranteedThroughputProbabilisticSampler(%s, %s, %s)' \
@@ -464,7 +473,10 @@ class RemoteControlledSampler(Sampler):
         elif s_type == SamplingManager.SamplingStrategyType.RATE_LIMITING:
             mtps = response[RATE_LIMITING_SAMPLING_STR][MAX_TRACES_PER_SECOND_STR]
             if 0 <= mtps < 500:
-                self.sampler = RateLimitingSampler(max_traces_per_second=mtps)
+                if isinstance(self.sampler, RateLimitingSampler):
+                    self.sampler.update(max_traces_per_second=mtps)
+                else:
+                    self.sampler = RateLimitingSampler(max_traces_per_second=mtps)
             else:
                 raise ValueError(
                     'Rate limiting parameter not in [0, 500] range: %s' % mtps)
--- a/jaeger_client/sampler.py
+++ b/jaeger_client/sampler.py
@@ -153,9 +153,6 @@ class RateLimitingSampler(Sampler):
                 SAMPLER_PARAM_TAG_KEY: max_traces_per_second,
             }
         )
-        self._init_rate_limiter(max_traces_per_second)
-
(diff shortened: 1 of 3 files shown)


tests/test_sampler.py
```diff
@@ -314,28 +313,30 @@ def test_sampling_request_callback():
...
     # noinspection PyProtectedMember
     sampler._sampling_request_callback(return_value)
-    assert '%s' % sampler.sampler == 'ProbabilisticSampler(0.002)'
+    assert '%s' % sampler.sampler == 'ProbabilisticSampler(0.002)', 'sampler should have changed to probabilistic'
 
-    # Strategy hasn't changed
     # noinspection PyProtectedMember
     sampler._sampling_request_callback(return_value)
-    assert '%s' % sampler.sampler == 'ProbabilisticSampler(0.002)'
+    assert '%s' % sampler.sampler == 'ProbabilisticSampler(0.002)', "strategy hasn't changed so sampler should not change"
```
yurishkuro: should this test be comparing the before/after sampler references?

jaeger_client/sampler.py
```diff
@@ -421,56 +421,64 @@ def _sampling_request_callback(self, future):
...
-    def close(self):
+    def _update_sampler(self, response):
         with self.lock:
-            self.running = False
-            if self.periodic is not None:
-                self.periodic.stop()
+            if response.get(OPERATION_SAMPLING_STR):
+                self._update_adaptive_sampler(response.get(OPERATION_SAMPLING_STR))
+            else:
+                try:
```
yurishkuro: why not include adaptive in try/catch?

jaeger_client/sampler.py
```diff
@@ -192,6 +195,13 @@ def __eq__(self, other):
         d1['last_tick'] = d2['last_tick']
         return d1 == d2
 
+    def update(self, max_traces_per_second):
```
yurishkuro: whaaaa? The whole point of using sampler.Equals was to solve that exact issue.

jaeger_client/sampler.py
yurishkuro: it doesn't know, but might in the future. Instead of catching exceptions at multiple levels I would rather catch them once in _update_sampler, to ensure that this thread doesn't die.

jaeger_client/sampler.py
black-adder: i dun gon goofed, still the update logic doesn't check for equality so I'll make that change

jaeger_client/sampler.py
```diff
@@ -421,56 +421,65 @@ def _sampling_request_callback(self, future):
...
-        operation_strategies = response.get(OPERATION_SAMPLING_STR)
-        if operation_strategies:
-            if isinstance(sampler, AdaptiveSampler):
-                return sampler, operation_strategies
-            return AdaptiveSampler(operation_strategies, self.max_operations), operation_strategies
         if s_type == SamplingManager.SamplingStrategyType.PROBABILISTIC:
             sampling_rate = response[PROBABILISTIC_SAMPLING_STR][SAMPLING_RATE_STR]
             if 0 <= sampling_rate <= 1.0:
-                return ProbabilisticSampler(rate=sampling_rate), None
+                self.sampler = ProbabilisticSampler(rate=sampling_rate)
```
yurishkuro: you can define `new_sampler` variable outside of `if` statement and do the `__eq__` test for all types. Otherwise you are forcing the assumption that replacing probabilistic sampling is safe - while it may be correct assumption, it's not necessary to force. Updating the sampler in one place in this function simplifies the logic.

jaeger_client/sampler.py
```diff
@@ -421,56 +421,65 @@ def _sampling_request_callback(self, future):
...
             else:
                 raise ValueError(
                     'Probabilistic sampling rate not in [0, 1] range: %s' % sampling_rate)
         elif s_type == SamplingManager.SamplingStrategyType.RATE_LIMITING:
             mtps = response[RATE_LIMITING_SAMPLING_STR][MAX_TRACES_PER_SECOND_STR]
             if 0 <= mtps < 500:
-                return RateLimitingSampler(max_traces_per_second=mtps), None
+                new_sampler = RateLimitingSampler(max_traces_per_second=mtps)
+                if not self.sampler.__eq__(new_sampler):
```
yurishkuro: why use `_eq__` and not `!=`? We have unit tests that prove `!=` works correctly.

tests/test_sampler.py
```diff
@@ -314,28 +313,31 @@ def test_sampling_request_callback():
...
     sampler._sampling_request_callback(return_value)
-    assert '%s' % sampler.sampler == 'ProbabilisticSampler(0.002)'
+    assert '%s' % sampler.sampler == 'ProbabilisticSampler(0.002)', 'sampler should have changed to probabilistic'
+    prev_sampler = sampler.sampler
 
-    # Strategy hasn't changed
     # noinspection PyProtectedMember
     sampler._sampling_request_callback(return_value)
-    assert '%s' % sampler.sampler == 'ProbabilisticSampler(0.002)'
+    assert prev_sampler == sampler.sampler, "strategy hasn't changed so sampler should not change"
```
yurishkuro: this may not be the correct comparison, doesn't Python treat `==` as `__eq__`? We need to compare references.

tests/test_sampler.py
```diff
@@ -417,31 +450,36 @@ def test_parse_sampling_strategy():
...
+    remote_sampler._update_sampler(new_response)
+    assert '%s' % remote_sampler.sampler == 'AdaptiveSampler(0.51, 3, 10)'
 
-    sampler.close()
+    # noinspection PyProtectedMember
+    remote_sampler._update_sampler({"strategyType":0,"probabilisticSampling":{"samplingRate":0.004}})
+    assert '%s' % remote_sampler.sampler == 'ProbabilisticSampler(0.004)', \
+        'should not fail going from adaptive sampler to probabilistic sampler'
+
+    remote_sampler.close()
```
yurishkuro: nit: the last two tests are almost 100 lines each. Perhaps there's a simpler way to structure them? Can they be data driven?

tests/test_sampler.py
Subzidion: `x == y` calls `x.__eq__(y)`. [Source](https://docs.python.org/2/reference/datamodel.html#object.__eq__).

PR #39, opened 2017-03-27
Add defaults for all sampling strategies

black-adder:


jaeger_client/sampler.py
```diff
@@ -45,6 +45,7 @@
 SAMPLER_TYPE_TAG_KEY = 'sampler.type'
 SAMPLER_PARAM_TAG_KEY = 'sampler.param'
 DEFAULT_SAMPLING_PROBABILITY = 0.001
+DEFAULT_LOWER_BOUND = 0.0016666 # sample once every 10 minutes
```
yurishkuro: `1.0 / (10 * 60)` would be better value

jaeger_client/sampler.py
```diff
@@ -480,3 +482,21 @@ def close(self):
             self.running = False
             if self.periodic is not None:
                 self.periodic.stop()
+
+
+def get_sampling_probability(strategy=None):
+    if not strategy:
+        strategy = {}
```
yurishkuro: why not just return some default right away?

PR #42, opened 2017-04-27
Refactor out rate limiter

black-adder:


jaeger_client/rate_limiter.py
```diff
@@ -0,0 +1,65 @@
...
+
+import time
+
+
+class RateLimiter(object):
+    """
+    RateLimiter is based on leaky bucket algorithm, formulated in terms
+    of a credits balance that is replenished every time check_credit()
+    method is called (tick) by the amount proportional to the time
+    elapsed since the last tick, up to max of credits_per_second. A call
```
yurishkuro: up to the max balance

jaeger_client/sampler.py
```diff
@@ -157,38 +157,23 @@ def __init__(self, max_traces_per_second=10):
         assert max_traces_per_second >= 0, \
             'max_traces_per_second must not be negative'
         self.credits_per_second = max_traces_per_second
```
yurishkuro: I think it makes sense to keep the var name as traces per second

jaeger_client/sampler.py
```diff
@@ -157,38 +157,23 @@ def __init__(self, max_traces_per_second=10):
...
-        self.max_balance = max_traces_per_second
-        if self.max_balance < 1.0:
-            self.max_balance = 1.0
-        self.balance = self.max_balance
-        self.last_tick = self.timestamp()
-        self.item_cost = 1.0
+        self.rate_limiter = RateLimiter(
+            self.credits_per_second,
```
yurishkuro: prefer named params

jaeger_client/rate_limiter.py
```diff
@@ -0,0 +1,65 @@
...
+    by instantiating the Rate Limiter with the max number of messages a
+    service is allowed to emit per second, and calling check_credit(1.0)
+    for each message to determine if the message is within the rate limit.
+
+    It can also be used to limit the rate of traffic in bytes, by setting
+    credits_per_second to desired throughput as bytes/second, and calling
+    check_credit() with the actual message size.
+    """
+
+    def __init__(self, credits_per_second=10.0, max_balance=10.0):
```
yurishkuro: not sure defaults make sense for this class

PR #45, opened 2017-05-12
Add metrics factory and allow tags for metrics

black-adder:


jaeger_client/metrics.py
```diff
@@ -21,69 +21,75 @@
...
-    """Number of errors decoding tracing context"""
+        self._count = count
+        if not callable(self._count):
+            self._count = None
 
-    REPORTER_SUCCESS = '%s.spans.reported' % prefix
-    """Number of spans successfully reported"""
+    def increment(self, value):
+        if self._count:
+            self._count(self.name, value, self.tags)
```
black-adder: All these functions will need to be mapped to https://github.com/uber/clay/blob/master/clay/stats.py which don't have a tags parameter. Will probably have to update the internal python-client to directly use m3-stats?

jaeger_client/config.py
```diff
@@ -81,24 +81,23 @@ class Config(object):
     _initialized = False
     _initialized_lock = threading.Lock()
 
-    def __init__(self, config, metrics=None, service_name=None):
+    def __init__(self, config, metrics_factory=None, service_name=None):
```
yurishkuro: this is a breaking change, how do you suggest we roll it out internally? Is it possible to accept the old `metrics` param as deprecated and construct a new factory that delegates to the old class, as well as takes care of converting tags to a single string key (similar to how expvar factory injaeger-lib works)?

jaeger_client/metrics.py
```diff
@@ -21,69 +21,75 @@
...
+        tags_cpy = self.tags.copy()
+        tags_cpy.update(tags)
+        return tags_cpy
 
-    SPANS_SAMPLED = '%s.spans.sampled' % prefix
-    """Number of sampled spans started/finished by this tracer"""
 
-    SPANS_NOT_SAMPLED = '%s.spans.not-sampled' % prefix
-    """Number of not-sampled spans started/finished by this tracer"""
+class Counter(object):
```
yurishkuro: Have you considered having the factory just return functions, instead of objects? Given the lack of interfaces / compiler checks, functions might be simpler to deal with. It would also be simpler to wrap legacy Metrics as MetricsFactory.

jaeger_client/reporter.py
```diff
@@ -221,6 +222,18 @@ def _flush(self):
         yield self.queue.join()
 
 
+class ReporterMetrics:
+    def __init__(self, metrics_factory):
+        self.reporter_success = \
+            metrics_factory.counter(name='jaeger.spans', tags={'reported': 'True'})
```
yurishkuro: you probably want lower-case true/false, to match Go metrics.

jaeger_client/reporter.py
```diff
@@ -221,6 +222,18 @@ def _flush(self):
...
+        self.reporter_failure = \
+            metrics_factory.counter(name='jaeger.spans', tags={'reported': 'False'})
+        self.reporter_dropped = \
+            metrics_factory.counter(name='jaeger.spans', tags={'dropped': 'True'})
```
yurishkuro: is this naming pattern consistent with Go?  It looks weird. E.g. there's never going to be dropped=false, so wouldn't a tag-less name `jaeger.spans.reporter-dropped` or such be more appropriate?

I would actually suggest starting a s/s and pulling all the names across 4 clients into it, and then coming up with a single consistent naming scheme (it can be a separate diff from this, if we don't release this one).

jaeger_client/metrics.py
```diff
@@ -21,69 +21,155 @@
...
+        self._count = count
+        self._gauge = gauge
+        self._timing = timing
+        self._tags = tags
+        if not callable(self._count):
+            self._count = None
+        if not callable(self._gauge):
+            self._gauge = None
+        if not callable(self._timing):
+            self._timing = None
```
yurishkuro: what does this achieve? It seems that instead you can default to no-op functions and change the public methods to simply return the `_` values, it would simplify all the derived classes. Or leave the defaulting to the Noop class

black-adder: I'm not following this 100%, I defaulted to no-op functions but changing the public method on the base class to return _ doesn't make sense to me given that this means the API will be different for the BaseClass vs the Subclasses

jaeger_client/metrics.py
```diff
@@ -21,69 +21,155 @@
...
+
+    def _merge_tags(self, tags=None):
+        if not self._tags:
+            return tags
+        tags_cpy = self._tags.copy()
+        tags_cpy.update(tags)
+        return tags_cpy
+
+
+class NoopMetricsFactory(MetricsFactory):
```
yurishkuro: I don't understand why you need this class. If you change the base class to actually invoke the functions it gets as constructor arguments from the main methods (instead of raising not implemented), then the base class becomes Noop (especially since you default to _noop it it already).

black-adder: The thing is the functions we get from the constructor as arguments have different signatures. Our default count function looks like this: `count(self, key, value)` but our m3 count function looks like: `count(self, key, n=1, sample=1.0, tags={}, include_deployment=False)`

Evoking the default count method in the base class is useless without the correct parameters so I decided to throw a notimplemented error instead of calling the count function and created a noop subclass that says it explicitly noops instead.

yurishkuro: * `count` function is a bad name - this is a factory function, so should be called `counter` or even `createCounter` to be explicit. It returns a single-argument function `def increment(value)`. This by default it can `return _noop`.

jaeger_client/metrics.py
```diff
@@ -21,69 +21,143 @@
...
+        Generates a new counter from the given name and tags and returns
+        a callable function used to increment the counter.
+        :param name: name of the counter
+        :param tags: tags for the counter
+        :return: a callable function which takes the value to increase
+        the counter by ie. def increment(value)
+        """
+        def increment(value):
+            return self._noop()
+        return increment
```
yurishkuro: why can't this be simply `return self._noop`? And if it can, can't we just throw away all of the constructor code?  Then in the Legacy class, instead of populating these fields, just store `self._metrics` class and use it directly in `create_timer` and such.

jaeger_client/metrics.py
```diff
@@ -21,69 +21,122 @@
...
 
-class Metrics(object):
-    """Provides an abstraction of metrics reporting framework."""
+class MetricsFactory(object):
+    """Generates new metrics."""
 
-    # TODO do we want to automatically include host name?
-    prefix = 'jaeger'
+    def __init__(self, tags=None):
+        self._tags = tags
```
yurishkuro: not used - don't need to store

jaeger_client/metrics.py
```diff
@@ -21,69 +21,122 @@
...
-    TRACES_JOINED_SAMPLED = '%s.traces-joined.sampled' % prefix
-    """Number of externally started sampled traces this tracer joined"""
+    def create_timer(self, name, tags=None):
+        """
+        Generates a new timer from the given name and tags and returns
+        a callable function used to record a duration.
+        :param name: name of the timer
+        :param tags: tags for the timer
+        :return: a callable function which takes the duration to
+        record ie. def record(duration)
```
yurishkuro: what is the measurement of duration, seconds as a float value?

jaeger_client/metrics.py
```diff
@@ -21,69 +21,122 @@
...
-    """Number of times the Sampler failed to retrieve sampling strategy"""
+    def create_gauge(self, name, tags=None):
+        key = self._get_key(name, self._merge_tags(tags))
 
-    def __init__(self, count=None, gauge=None):
+        def update(value):
+            return self._metrics.gauge(key, value)
+        return update
+
+    def _get_key(self, name, tags=None):
```
yurishkuro: I think `_merge_tags` and `_get_key` belong to the same class.

jaeger_client/metrics.py
```diff
@@ -21,69 +21,122 @@
...
+        if not tags:
+            return name
+        key = name
+        for k in sorted(tags.iterkeys()):
+            key = key + '|' + str(k) + '=' + str(tags[k])
```
yurishkuro: not sure this is how we want to make the key anymore. In the expvar bridge I had to do a similar toString for the tags, and the format I used was `{name}.{tag-key}_{tag-value}..`, with tags sorted by keys. This makes it relatively friendly to grafana path-based functions, and avoids `|` and `=` which might be reserved characters in some metrics systems.

jaeger_client/metrics.py
black-adder: _merge_tags() will be used by both LegacyMetricsFactory and M3MetricsFactory so I put it in the base class and added tags as a parameter to the base class. I guess it's all possible to make _merge_tags static but I think this is cleaner. Also, _get_key will only be used by LegacyMetricsFactory so think it's fine as is.

jaeger_client/metrics.py
black-adder: so clay uses milliseconds and m3 used microseconds, ill make it microseconds and cast it for clay (legacy)

jaeger_client/config.py
```diff
@@ -81,24 +81,23 @@ class Config(object):
     _initialized = False
     _initialized_lock = threading.Lock()
 
-    def __init__(self, config, metrics=None, service_name=None):
+    def __init__(self, config, metrics=None, service_name=None, metrics_factory=None):
         """
         :param metrics: an instance of Metrics class, or None
```
yurishkuro: describe new param, mark old as deprecated

jaeger_client/metrics.py
```diff
@@ -21,69 +21,128 @@
...
+        tags_cpy.update(tags)
+        return tags_cpy
 
-    SPANS_NOT_SAMPLED = '%s.spans.not-sampled' % prefix
-    """Number of not-sampled spans started/finished by this tracer"""
 
-    TRACER_DECODING_ERRORS = '%s.decoding.errors' % prefix
-    """Number of errors decoding tracing context"""
+class LegacyMetricsFactory(MetricsFactory):
+    """A MetricsFactory wrapper around Metrics."""
```
yurishkuro: A MetricsFactory adapter for legacy Metrics class.

jaeger_client/metrics.py
```diff
@@ -21,69 +21,128 @@
...
 
-    REPORTER_SUCCESS = '%s.spans.reported' % prefix
-    """Number of spans successfully reported"""
+    def __init__(self, metrics, tags=None):
```
yurishkuro: who would ever pass tags to this class? Isn't its use completely internal?

jaeger_client/metrics.py
```diff
@@ -21,69 +21,128 @@
...
+            tag_separator = '_'
+            if first_tag:
+                tag_separator = '.'
+                first_tag = False
+            key = key + tag_separator + str(k) + '-' + str(tags[k])
+        return key
+
+
+class Metrics(object):
+    """Provides an abstraction of metrics reporting framework."""
```
yurishkuro: say something about legacy / /deprecated

jaeger_client/metrics.py
```diff
@@ -21,69 +21,128 @@
...
+
+    def __init__(self, count=None, gauge=None, timing=None):
         """
         :param count: function (key, value) to emit counters
         :param gauge: function (key, value) to emit gauges
+        :param timing: function (key, value) to emit timings
```
yurishkuro: add "value in milliseconds"

jaeger_client/reporter.py
```diff
@@ -79,7 +79,7 @@ class Reporter(NullReporter):
     """Receives completed spans from Tracer and submits them out of process."""
     def __init__(self, channel, queue_capacity=100, batch_size=10,
                  flush_interval=DEFAULT_FLUSH_INTERVAL, io_loop=None,
-                 error_reporter=None, metrics=None, **kwargs):
+                 error_reporter=None, metrics_factory=None, **kwargs):
```
yurishkuro: this is not backwards compatible. Reporter is a public class, so better to support the old `metrics` kv as well.

jaeger_client/reporter.py
```diff
@@ -89,7 +89,7 @@ def __init__(self, channel, queue_capacity=100, batch_size=10,
         :param io_loop: which IOLoop to use. If None, try to get it from
             channel (only works if channel is tchannel.sync)
         :param error_reporter:
-        :param metrics:
+        :param metrics_factory:
```
yurishkuro: both + mention that `metrics` is deprecated

jaeger_client/sampler.py
```diff
@@ -327,7 +327,7 @@ def __init__(self, channel, service_name, **kwargs):
             - sampling_refresh_interval: interval in seconds for polling
               for new strategy
             - logger:
-            - metrics: metrics facade, used to emit metrics on errors
+            - metrics_factory: used to generate metrics for errors
```
yurishkuro: probably the same thing, support legacy `metrics` too

jaeger_client/tracer.py
```diff
@@ -35,14 +35,14 @@
...
-from .metrics import Metrics
+from .metrics import MetricsFactory
 from .utils import local_ip
 
 logger = logging.getLogger('jaeger_tracing')
 
 
 class Tracer(opentracing.Tracer):
-    def __init__(self, service_name, reporter, sampler, metrics=None,
+    def __init__(self, service_name, reporter, sampler, metrics_factory=None,
```
yurishkuro: definitely support legacy `metrics`

jaeger_client/metrics.py
```diff
@@ -21,69 +21,120 @@
...
+        if not tags:
+            return name
+        key = name
+        first_tag = True
+        for k in sorted(tags.iterkeys()):
+            tag_separator = '_'
+            if first_tag:
+                tag_separator = '.'
+                first_tag = False
+            key = key + tag_separator + str(k) + '-' + str(tags[k])
```
yurishkuro: that's still different from expvar adapter. I suggest always using `.` as separator between name and tags, including between two tags, and `_` as separator between key and value. E.g. 

`operation-counts.proto_http.status_2xx`

jaeger_client/utils.py
```diff
@@ -22,25 +22,19 @@
...
-
 
 class ErrorReporter(object):
     """
     Reports errors by emitting metrics, and if logger is provided,
     logging the error message once every log_interval_minutes
     """
 
-    def __init__(self, metrics, logger=None, log_interval_minutes=15):
+    def __init__(self, logger=None, log_interval_minutes=15):
```
yurishkuro: please keep the metrics argument. For some reason we allowed reporter to be created externally, so it's a public API now.

jaeger_client/utils.py
```diff
@@ -22,25 +22,21 @@
...
 
 class ErrorReporter(object):
     """
     Reports errors by emitting metrics, and if logger is provided,
     logging the error message once every log_interval_minutes
+
+    N.B. metrics will be deprecated in the future
     """
 
     def __init__(self, metrics, logger=None, log_interval_minutes=15):
```
yurishkuro: I suggest `metrics=None` - keep it for backwards compatibility, but don't require it.

jaeger_client/metrics.py
```diff
@@ -21,69 +21,115 @@
...
+        return update
 
-    REPORTER_QUEUE_LENGTH = '%s.queue_length' % prefix
-    """Current number of spans in the reporter queue"""
+    def _get_key(self, name, tags=None):
+        if not tags:
+            return name
+        key = name
+        for k in sorted(tags.iterkeys()):
+            key = key + '.' + str(k) + '_' + str(tags[k])
```
yurishkuro: we need to normalize the values to a set of strings that are safe for metric name, because operation names can be anything. See how it's done in the Go client (doesn't have to be as flexible as there). And let's add tests to check that operation names are properly sanitized.

black-adder: next PR, I have it as a part of the rpc metrics stuff

jaeger_client/utils.py
black-adder: issue is we have some places where we call ErrorReporter(metrics, logger=logger) which would break if we made metrics None. With the change you suggest, it would be a breaking change.

jaeger_client/metrics.py
yurishkuro: ok

jaeger_client/utils.py
yurishkuro: I don't follow. `ErrorReporter(metrics, logger=logger)` is not in conflict with metrics param having a default value. We can change all internal calls not to pass metrics (like you did in the recent commit where you always pass `Metrics()`), but the public API is still able to accept the metrics object.

yurishkuro: anyway, this is not a blocker, I will go the last review and we can merge

PR #54, opened 2017-07-05
Save baggage in span

black-adder:


jaeger_client/span.py
```diff
@@ -139,9 +139,20 @@ def log_kv(self, key_values, timestamp=None):
         return self
 
     def set_baggage_item(self, key, value):
+        prev_value = self.get_baggage_item(key=key)
         new_context = self.context.with_baggage_item(key=key, value=value)
         with self.update_lock:
             self._context = new_context
+        if self.is_sampled():
+            logs = {
+                'event': 'baggage',
```
oibe: should 'baggage' be a constant in a file?

PR #57, opened 2017-07-08
Futurize for Py3 

MrSaints:
As per the suggestion https://github.com/uber/jaeger-client-python/pull/43#issuecomment-308895785, this PR runs `futurize` against the source with small fixes to ensure the Python 2 tests continues to pass. I thought I make this PR to get Python 3 support moving (even if it is in the smallest way possible).

### DIFF
--- a/jaeger_client/sampler.py
+++ b/jaeger_client/sampler.py
@@ -21,9 +21,10 @@
 from __future__ import absolute_import
 from __future__ import division
 from builtins import object
+import json
 import logging
 import random
-import json
+import six
 
 from threading import Lock
 from tornado.ioloop import PeriodicCallback
@@ -308,7 +309,7 @@ class AdaptiveSampler(Sampler):
                 ProbabilisticSampler(self.default_sampling_probability)
 
     def close(self):
-        for _, sampler in self.samplers.items():
+        for _, sampler in six.iteritems(self.samplers):
             sampler.close()
 
     def __str__(self):
--- a/jaeger_client/sampler.py
+++ b/jaeger_client/sampler.py
@@ -19,6 +19,9 @@
 # THE SOFTWARE.
 
 from __future__ import absolute_import
+from __future__ import division
+from builtins import object
+from past.utils import old_div
 import logging
 import random
 import json
@@ -45,7 +48,7 @@ default_logger = logging.getLogger('jaeger_tracing')
 SAMPLER_TYPE_TAG_KEY = 'sampler.type'
 SAMPLER_PARAM_TAG_KEY = 'sampler.param'
 DEFAULT_SAMPLING_PROBABILITY = 0.001
-DEFAULT_LOWER_BOUND = 1.0 / (10.0 * 60.0)  # sample once every 10 minutes
+DEFAULT_LOWER_BOUND = old_div(1.0, (10.0 * 60.0))  # sample once every 10 minutes
 DEFAULT_MAX_OPERATIONS = 2000
 
 STRATEGIES_STR = 'perOperationStrategies'
@@ -306,7 +309,7 @@ class AdaptiveSampler(Sampler):
                 ProbabilisticSampler(self.default_sampling_probability)
 
     def close(self):
-        for _, sampler in self.samplers.iteritems():
+        for _, sampler in self.samplers.items():
             sampler.close()
 
     def __str__(self):
--- a/jaeger_client/sampler.py
+++ b/jaeger_client/sampler.py
@@ -21,7 +21,6 @@
 from __future__ import absolute_import
 from __future__ import division
 from builtins import object
-from past.utils import old_div
 import logging
 import random
 import json
@@ -48,7 +47,7 @@ default_logger = logging.getLogger('jaeger_tracing')
 SAMPLER_TYPE_TAG_KEY = 'sampler.type'
 SAMPLER_PARAM_TAG_KEY = 'sampler.param'
 DEFAULT_SAMPLING_PROBABILITY = 0.001
-DEFAULT_LOWER_BOUND = old_div(1.0, (10.0 * 60.0))  # sample once every 10 minutes
+DEFAULT_LOWER_BOUND = 1.0 / (10.0 * 60.0)  # sample once every 10 minutes
 DEFAULT_MAX_OPERATIONS = 2000
 
 STRATEGIES_STR = 'perOperationStrategies'
--- a/jaeger_client/thrift.py
+++ b/jaeger_client/thrift.py
@@ -18,6 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+from past.builtins import basestring
 import socket
 import struct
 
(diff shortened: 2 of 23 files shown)


yurishkuro:
Thanks for this PR, @MrSaints!

This change looks good, mostly. The two concerns I have are:
  1. The use of `old_div` seems unnecessary. There aren't that many occurrences, can we replace them with `/` for floats and `//` for ints? It is usually apparent from the context which numbers we're dealing with.
  1. The change from `iteritems()/iterkeys()` to `items()/keys()` is not good, since in Py2 the latter will cause creation of new arrays. Can we replace all instances with respective proxy methods from `six`?

And there are some lint errors in the build:
```
flake8 jaeger_client crossdock tests
jaeger_client/codecs.py:25:1: E402 module level import not at top of file
jaeger_client/codecs.py:26:1: E402 module level import not at top of file
jaeger_client/codecs.py:27:1: E402 module level import not at top of file
jaeger_client/codecs.py:27:22: E401 multiple imports on one line
jaeger_client/codecs.py:29:1: E402 module level import not at top of file
jaeger_client/codecs.py:33:1: E402 module level import not at top of file
jaeger_client/codecs.py:38:1: E402 module level import not at top of file
jaeger_client/metrics.py:28:1: E302 expected 2 blank lines, found 0
make: *** [lint] Error 1
```

MrSaints:
Cheers for the quick feedback @yurishkuro. I've addressed them in the new commits.

jaeger_client/sampler.py
```diff
@@ -19,9 +19,12 @@
 # THE SOFTWARE.
 
 from __future__ import absolute_import
+from __future__ import division
```
yurishkuro: Q: is this import necessary?

tests/test_sampler.py
```diff
@@ -198,7 +201,7 @@ def test_adaptive_sampler():
     sampled, tags = sampler.is_sampled(MAX_INT-10, "new_op")
     assert sampled
     assert tags == get_tags('probabilistic', 0.51)
-    sampled, tags = sampler.is_sampled(MAX_INT+(MAX_INT/4), "new_op")
+    sampled, tags = sampler.is_sampled(MAX_INT+(old_div(MAX_INT,4)), "new_op")
```
yurishkuro: can we change back to `/`?

tests/test_sampler.py
```diff
@@ -207,7 +210,7 @@ def test_adaptive_sampler():
     sampled, tags = sampler.is_sampled(MAX_INT-10, "new_op_2")
     assert sampled
     assert tags == get_tags('probabilistic', 0.51)
-    sampled, _ = sampler.is_sampled(MAX_INT+(MAX_INT/4), "new_op_2")
+    sampled, _ = sampler.is_sampled(MAX_INT+(old_div(MAX_INT,4)), "new_op_2")
```
yurishkuro: can we change back to `/`?

tests/test_sampler.py
```diff
@@ -18,6 +19,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+from builtins import range
+from past.utils import old_div
```
yurishkuro: can we remove this?

tests/test_sampler.py
```diff
@@ -167,7 +170,7 @@ def test_guaranteed_throughput_probabilistic_sampler():
     sampled, tags = sampler.is_sampled(MAX_INT-10)
     assert sampled
     assert tags == get_tags('probabilistic', 0.51)
-    sampled, tags = sampler.is_sampled(MAX_INT+(MAX_INT/4))
+    sampled, tags = sampler.is_sampled(MAX_INT+(old_div(MAX_INT,4)))
```
yurishkuro: can we use `/`?

jaeger_client/sampler.py
MrSaints: I'm not too sure. Seems like it.

```
The future division statement, spelled from __future__ import division , will change the / operator to mean true division throughout the module.
```

jaeger_client/thrift.py
```diff
@@ -117,7 +118,7 @@ def timestamp_micros(ts):
     :param ts:
     :return:
     """
-    return long(ts * 1000000)
+    return int(ts * 1000000)
```
MrSaints: We can probably switch between `int`, and `long` using `six`.

tests/test_sampler.py
MrSaints: Done :)

MrSaints:
I've made the changes @yurishkuro, but I'm not too sure what's going on with one of the Travis `COVER=1` test: `INTERNALERROR> RuntimeError: dictionary changed size during iteration`

black-adder:
The code coverage issue looks related to: https://bitbucket.org/ned/coveragepy/issues/581/44b1-44-breaking-in-ci we probably want to use an older version of coverage until it's fixed

jaeger_client/thrift.py
black-adder: do we need to make that change for this PR?

MrSaints: I've made that change, I'm unsure if the `int` behaves as expected for py3, but we can do that in a different PR.

yurishkuro:
> `long_fn = int if six.PY3 else long`

@MrSaints a couple questions:
  * Will ^^ actually parse/compile with Py3?  I thought 3.x decommissioned `long`.
  * from reading the [PEP 237](https://www.python.org/dev/peps/pep-0237/) and [SO question](https://stackoverflow.com/questions/2104884/how-does-python-manage-int-and-long) it seems the overflow on ints has been eliminated in favor of automatic upgrading of the value from a machine word to a malloc-ed memory, making this transparent to the end user. I don't know if we can find a 32bit machine to test this, any thoughts?

Even on 64bit machine the automatic type conversion can be seen:
```python
>>> import sys;
>>> sys.maxint
9223372036854775807
>>> type(sys.maxint)
<type 'int'>
>>> type(sys.maxint + 1)
<type 'long'>
>>> type(sys.maxint * 100)
<type 'long'>
>>> (type(sys.maxint * 100), sys.maxint * 100)
(<type 'long'>, 922337203685477580700L)
```

i.e. an `int` value times 100 is auto-converted to `long`.

MrSaints:
(1) Whoops @yurishkuro. I've cleaned it up. 
(2) We can do more vigorous testing with Python 3. I believe it should be fine if it does not change the existing behaviour in Python 2.

yurishkuro:
Attempting to fix the coverage issue here #61

yurishkuro:
@MrSaints can you rebase or merge from master to pick up #61?

MrSaints:
@yurishkuro Cheers for the fix. And rebased.

MrSaints:
We good to merge?

MrSaints:
Cheers for all the feedback, and for pushing this out @yurishkuro @black-adder

Groxx:
Does this `install_aliases()` call taint the entire environment that imports it, not just this source file?  If yes, it seems like this may be breaking anything expecting e.g. a specific `html` library (or a specific version), like https://pypi.python.org/pypi/html

olevold:
I still get  `ImportError: No module named 'ttypes'`

MrSaints:
Are you running on Python 3 @olevold ? There's still work that needs to be done (e.g. on Thrift side) to get it working on Python 3.

olevold:
Ok, I thought it had been solved. Sorry

PR #63, opened 2017-07-10
Organize imports

black-adder:
Using https://github.com/timothycrosley/isort to automatically organize our imports

### DIFF
--- a/jaeger_client/metrics/prometheus.py
+++ b/jaeger_client/metrics/prometheus.py
@@ -12,9 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from jaeger_client.metrics import MetricsFactory
 from collections import defaultdict
+
 from prometheus_client import Counter, Gauge
+
+from jaeger_client.metrics import MetricsFactory
 
 
 class PrometheusMetricsFactory(MetricsFactory):
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,7 @@ setup(
             'tchannel>=0.27;python_version<"3"',
             'opentracing_instrumentation>=2,<3',
             'prometheus_client',
+            'isort<4.3',
         ]
     },
 )
--- a/tests/test_TUDPTransport.py
+++ b/tests/test_TUDPTransport.py
@@ -14,8 +14,9 @@
 
 from __future__ import absolute_import
 
-import mock
 import unittest
+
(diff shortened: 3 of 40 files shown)


yurishkuro:
@black-adder are you sure we want this? seems like it's not worth the noise

black-adder:
I hate the imports being all over the place. I don't think it's too much noise tbf, just a one time thing and then it's handled automatically from then on.

yurishkuro:
I am not opposed to doing that if you want to fix the PR. But does this tool support some sort of formatting style? E.g. this change is making imports less readable imo:

https://github.com/jaegertracing/jaeger-client-python/pull/63/files#diff-4f42106af9829128b1247b591a945a22L5

black-adder:
ill take a look see

yurishkuro:
Fails in py3 build with
```
ERROR: /home/travis/build/jaegertracing/jaeger-client-python/tests/test_crossdock.py Imports are incorrectly sorted.
```

I think running lint in each Py version is a bad idea as there can be incompatibilities.

jaeger_client/metrics/prometheus.py
```diff
@@ -12,10 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from jaeger_client.metrics import MetricsFactory
 from collections import defaultdict
+
 from prometheus_client import Counter, Gauge
 
+from jaeger_client.metrics import MetricsFactory
```
yurishkuro: probably would be better to import `from .`

setup.py
```diff
@@ -63,6 +63,7 @@
             'tchannel>=0.27;python_version<"3"',
             'opentracing_instrumentation>=2,<3',
             'prometheus_client',
+            'isort<4.3',
```
yurishkuro: why lower bound on minor?

tests/test_TUDPTransport.py
```diff
@@ -14,9 +14,10 @@
 
 from __future__ import absolute_import
 
-import mock
 import unittest
 
+import mock
```
yurishkuro: I don't get it. Are there actually some rules it uses? Like built-in modules first? That might explain the break on py3

yurishkuro:
closing for inactivity - given how old this is it would need to be re-done anyway.

PR #78, opened 2017-09-25
Allow tags to be passed to tracer constructor.

isaachier:


yurishkuro:
let's add tests. Also, most often the tracer is created via [config](https://github.com/uber/jaeger-client-python/blob/master/jaeger_client/config.py), so makes sense to add an option there as well.

btw, if you have time to find our why the build in Python 3 started failing, it would be good to fix (but if looks difficult, it can wait)

PR #81, opened 2017-09-26
Test creating tracer from config.

isaachier:


tests/test_config.py
```diff
@@ -42,6 +42,7 @@ def test_tags(self):
         os.environ['JAEGER_TAGS'] = 'a=b,c=d'
         c = Config({'tags': {'e': 'f'}}, service_name='x')
         assert c.tags == {'a': 'b', 'c': 'd', 'e': 'f'}
+        c.initialize_tracer()
```
yurishkuro: do you need to close it? It will start a thread. Or perhaps configure a noop reporter

isaachier: Didn't think about that. Good idea.

Issue #59, opened 2017-07-09
Support Python 3.x


yurishkuro:
Related: #41, #43, #57

MrSaints:
Remaining steps (in order of severity):

- [ ] Upgrade Thrift for Python 3 support _(I believe this is not going to be a backwards compatible change)_
- [ ] Fix `make test` errors [0]
- [ ] `opentracing_instrumentation` is not Python 3 compatible, and it is a `tests` requirement (`ModuleNotFoundError: No module named 'urlparse'`)

[0] Some test errors / failures that needs resolving:

- `opentracing.propagation.InvalidCarrierException: carrier not a collection`
- ` TypeError: '<=' not supported between instances of 'MagicMock' and 'MagicMock'`
- `TypeError: '>' not supported between instances of 'NoneType' and 'int'`
- Several others ...

yurishkuro:
As far as Thrift upgrade, we currently require unbound `thrift`. If possible, I would like to keep it this way, and install different versions in the Travis matrix build. If we need to regenerate the files under `jaeger_client/thrift_gen`, then we may need to create subdirs for different Thrift versions.

The main reason for this is that many services at Uber pin Thrift to older versions, and pip-compile won't be able to resolve it if we require `thrift>=0.10`

Re `opentracing_instrumentation` - it would be good to make it Py3 compatible as well. I don't think it uses Jaeger directly, it should only use `basic-tracer` which should be Py3 compatible.

dsludwig:
I've got a collection of patches that I used to make `jaeger-client 3.4.0` and `opentracing_instrumentation 2.2.0` run on Python 3, attached here. A number of these were covered by #57, but there are a few that are not. I was only trying to support Python 3 rather than both, so they're not applicable as-is, but hopefully this helps!

```diff
diff --git jaeger_client/codecs.py jaeger_client/codecs.py
index d25eb69..b4fca65 100644
--- jaeger_client/codecs.py
+++ jaeger_client/codecs.py
@@ -20,7 +20,7 @@
 
 from __future__ import absolute_import
 
-import urllib
+import urllib.parse
 
 from opentracing import (
     InvalidCarrierException,
@@ -64,29 +64,29 @@ class TextCodec(Codec):
             parent_id=span_context.parent_id, flags=span_context.flags)
         baggage = span_context.baggage
         if baggage:
-            for key, value in baggage.iteritems():
+            for key, value in baggage.items():
                 if self.url_encoding:
[...]

PR #94, opened 2017-11-06
new thrift files

MaxPresman:
Related to (https://github.com/jaegertracing/jaeger-idl/pull/37), this brings the new thrift files for CI runs.

Signed-off-by: Max Presman <max@pubnub.com>


MaxPresman:
@yurishkuro I understand. I guess the only way forward is It keep two sets of generated files? 

The lack of python3 is blocking adoption on using Jaeger : (. 

Is python 3 a priority for the project?

yurishkuro:
Right now we do not restrict `thrift` to 0.9.3, the dependency [is declared simply as `thrift`](https://github.com/jaegertracing/jaeger-client-python/blob/master/setup.py#L41).

To confirm, this works fine on master:
```
pip install thrift==0.10.0
make test
```

So I think we don't need to regenerate the source files in order to use 0.10. 

The reverse doesn't work, i.e. if I generate files with 0.10.0 then tests don't run with thrift 0.9.3 installed (`ImportError: cannot import name TFrozenDict`). So there might be some performance implication when using 0.10.0 module with 0.9.3-generated files.

If the mix and match doesn't work, then the fallback would be generating two sets of files and switching dynamically (started in #95). But right now it doesn't look like that's necessary.

> Is python 3 a priority for the project?

Different orgs (and maintainers) have different priorities. I **would** like to get the client compatible with py3, but we currently have many higher priority tasks at Uber, that's why the issue #59 is tagged "help wanted". A bunch of changes have already been done towards supporting py3, including fully supporting it in `opentracing_instrumentation` module that's a dependency for tests. 

If you have the cycles to finish what's remaining in #59, it would be https://youtu.be/8r7OR22sE2g.

MaxPresman:
@yurishkuro thanks for the responses; i'll close for now.

PR #96, opened 2017-11-13
Test py36

nehaljwani:


yurishkuro:
There are still some changes required for Py3, see #59

PR #99, opened 2017-11-21
fix(python3): add support. skindeep changes for #59

dudymas:
This has been lightly tested, and will hopefully foster discussion. I'm doing a spike for adding tracing to our team's django app with Jaeger and Python3 .

I'd like to merge this in, but please understand I'm only doing this to get around opentracing-contrib/python-django#7 . I have another PR (to another PR) for that issue:
opentracing-contrib/python-django#6
granduke/python-django#1 (this is the changes I've made to add python3 support and such)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,8 @@ thrift: idl-submodule thrift-image
 	${THRIFT} -o /data --gen py:${THRIFT_PY_ARGS} -out /data/$(THRIFT_GEN_DIR) /data/idl/thrift/zipkinCore.thrift
 	${THRIFT} -o /data --gen py:${THRIFT_PY_ARGS} -out /data/$(THRIFT_GEN_DIR) /data/idl/thrift/agent.thrift
 	${THRIFT} -o /data --gen py:${THRIFT_PY_ARGS} -out /data/$(THRIFT_GEN_DIR) /data/idl/thrift/sampling.thrift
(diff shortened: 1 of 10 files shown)


yurishkuro:
most of these changes are to auto-generated files. How did you do that, manually?

dudymas:
Yeah! ooph... I thought 0.9 wasn't working. I just checked, and 0.10.0 is available. one min while I test it out, then I'll update the branch. woah nelly... it changed a lot of the generated code!

dudymas:
Unfortunately, the 0.10.0 version of thrift seems to break functionality. And it also appears that it would require updates to the jaeger service...?
So that's a no go. The easiest solution, if you don't approve of the manual edits, would be to run a `sed` script after the thrift generation is done. Thoughts? That'd be easy to whip up...

yurishkuro:
iirc (based on #59) at runtime you can use Py3 with thrift-.10 while the classes are still generated by 0.9.3, but yes they do need to be fixed up a bit (see https://github.com/jaegertracing/jaeger-client-python/issues/59#issuecomment-317906782). If fixer-upper can be done with sed or otherwise automated then I am not against doing that.

dudymas:
Okay, see what you think. I found that sed on osx will break with linux unless I add a backup extension... thus I just went ahead and added that. Should work fine on all major three: mac/linux/wsl

yurishkuro:
lgtm. Could you please rebase off master so that the build is green?

dudymas:
rebased \o/ woohoo.

PR #101, opened 2017-11-27
Fixes for py3 support

nehaljwani:


yurishkuro:
most of these fixes either have been applied already or are not applicable.

PR #109, opened 2017-12-13
Encode unicode baggage keys/values to UTF-8

yurishkuro:
Replaces #108

There's also an outstanding bug report https://bugs.python.org/issue22231

The original issue is this exception:

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xb5 in position 295: ordinal not in range(128)
(10 additional frame(s) were not displayed)
...
  File "urllib3/connectionpool.py", line 356, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "python2.7/httplib.py", line 1052, in request
    self._send_request(method, url, body, headers)
  File "python2.7/httplib.py", line 1092, in _send_request
    self.endheaders(body)
  File "python2.7/httplib.py", line 1048, in endheaders
    self._send_output(message_body)
  File "python2.7/httplib.py", line 890, in _send_output
    msg += message_body

UnicodeDecodeError: 'ascii' codec can't decode byte 0xb5 in position 295: ordinal not in range(128)
```

### DIFF
--- a/tests/test_codecs.py
+++ b/tests/test_codecs.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import
 import unittest
 from collections import namedtuple
 import urllib2
+import six
 
 import mock
 import pytest
@@ -122,14 +123,14 @@ class TestCodecs(unittest.TestCase):
                     'trace-id': '100:7f:0:1',
                     'trace-attr-bender': 'Countess%20de%20la%20Roca',
                     'trace-attr-fry': 'Leela'}
+                for key, val in six.iteritems(carrier):
+                    assert isinstance(key, str)
+                    assert isinstance(val, str)
             else:
                 assert carrier == {
                     'trace-id': '100:7f:0:1',
                     'trace-attr-bender': 'Countess de la Roca',
                     'trace-attr-fry': 'Leela'}
-            for key, val in carrier.iteritems():
-                assert isinstance(key, str)
-                assert isinstance(val, str)
 
     def test_context_from_bad_readable_headers(self):
         codec = TextCodec(trace_id_header='Trace_ID',
@@ -326,6 +327,7 @@ def test_debug_id():
 
 
 def test_non_unicode_baggage(httpserver):
+    # httpserver is provided by pytest-localserver
     httpserver.serve_content(content='Hello', code=200, headers=None)
 
     tracer = Tracer(
--- a/tests/test_codecs.py
+++ b/tests/test_codecs.py
@@ -18,6 +18,7 @@ import unittest
 from collections import namedtuple
 import urllib2
 import six
+from future.types.newstr import newstr
 
 import mock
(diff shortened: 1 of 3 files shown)


debosmit:
Looking at this comment:

> utf-8 encoding is only one step in IRI encoding. Correct IRI encoding is non trivial and doesn't fall into the support policy for 2.7 (bug/security fixes). I think that the best that can be done for 2.7 is to enhance the documentation around HTTPConnection.__init__ (unicode hostnames should be IDNA-encoded with the built-in IDNA encoder) and HTTPConnection.request/putrequest noting that unicode paths should be IRI encoded, with a link to RFC 3987.

doesn't seem like the fix will ever make it to python 2.7.x either.

At the very best, seems like developers have to be aware of whatever headers they are injecting into the request, and making sure that they are `utf-8` encoded.

tests/test_codecs.py
```diff
@@ -111,7 +113,7 @@ def test_context_to_readable_headers(self):
             assert carrier == {'trace-id': '100:7f:0:1'}
 
             ctx._baggage = {
-                'fry': 'Leela',
+                'fry': u'Leela',
```
vprithvi: Could you also update the string to have a unicode character which doesn't have an ascii representation?

yurishkuro: there is a new test that does that

tests/test_codecs.py
```diff
@@ -319,3 +324,37 @@ def test_debug_id():
...
+        # don't sample to avoid logging baggage to the span
+        sampler=ConstSampler(False),
+    )
+    tracer.codecs[Format.TEXT_MAP] = TextCodec(url_encoding=True)
+
+    baggage = [
+        (b'key', b'value'),
+        (u'key', b'value'),
+        (b'key', bytes(chr(255))),
+        (u'caf\xe9', 'caf\xc3\xa9'),
```
vprithvi: The test name ` test_non_unicode_baggage` makes it seem like this tests string/byte baggage; but I don't see any assertions on the non unicode baggage items.

isaachier: The potential bug is in the httplib/urllib2 so IDK if there is a direct assertion we should do.

tests/test_codecs.py
vprithvi: I don't think that the new test, `test_non_ascii_baggage` verifies that the carrier carries the correctly encoded baggage string.

tests/test_codecs.py
```diff
@@ -319,3 +338,37 @@ def test_debug_id():
     tags = [t for t in span.tags if t.key == debug_header]
     assert len(tags) == 1
     assert tags[0].value == 'Coraline'
+
+
+def test_non_ascii_baggage(httpserver):
```
vprithvi: nit: this is really testing whether httplib works fine

PR #110, opened 2017-12-17
Increase max tag value length to 1024 and make it configurable

yurishkuro:
Resolves https://github.com/jaegertracing/jaeger/issues/508

### DIFF
--- a/jaeger_client/constants.py
+++ b/jaeger_client/constants.py
@@ -65,3 +65,6 @@ SAMPLER_TYPE_RATE_LIMITING = 'ratelimiting'
 # of traces per second.
 # noinspection SpellCheckingInspection
 SAMPLER_TYPE_LOWER_BOUND = 'lowerbound'
+
+# max length for tag values. Longer values will be truncated.
+MAX_TAG_VALUE_LENGTH = 1024
--- a/tests/test_reporter.py
+++ b/tests/test_reporter.py
@@ -225,7 +225,7 @@ class ReporterTest(AsyncTestCase):
 
     @gen_test
     def test_submit_batch_size_2(self):
-        reporter, sender = self._new_reporter(batch_size=2, flush=0.005)
+        reporter, sender = self._new_reporter(batch_size=2, flush=0.01)
         reporter.report_span(self._new_span('1'))
         yield tornado.gen.sleep(0.001)
         assert 0 == len(sender.futures)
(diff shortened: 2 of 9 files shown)


jaeger_client/constants.py
```diff
@@ -65,3 +65,6 @@
 # of traces per second.
 # noinspection SpellCheckingInspection
 SAMPLER_TYPE_LOWER_BOUND = 'lowerbound'
+
+# max length for tag values. Longer values will be truncated.
+MAX_TAG_VALUE_LENGTH = 1024
```
vprithvi: Why 1024? 
What is the purpose of curtailing the length? 
Is it better to have `MAX_TAG_VALUE_BYTES` instead? (Because unicode chars can span multiple bytes?)

tests/test_reporter.py
```diff
@@ -225,7 +225,7 @@ def test_submit_queue_full_batch_size_1(self):
 
     @gen_test
     def test_submit_batch_size_2(self):
-        reporter, sender = self._new_reporter(batch_size=2, flush=0.005)
+        reporter, sender = self._new_reporter(batch_size=2, flush=0.01)
```
vprithvi: Why is this change required in this diff?

tests/test_span.py
```diff
@@ -232,3 +232,12 @@ def test_span_component(tracer):
 
     span.set_tag(ext_tags.COMPONENT, 'crypt')
     assert span.component == 'crypt'
+
+
+def test_span_tag_value_max_length(tracer):
+    tracer.max_tag_value_length = 42
+    span = tracer.start_span(operation_name='x')
+    span.set_tag('x', 'x' * 50)
```
vprithvi: Could you add a test with unicode as well?

jaeger_client/constants.py
yurishkuro: python client does not have logic to calculate the size of the UDP packet, instead it just limits the number of spans per packet. Having unbounded tag/log values increases the risk of exceeding the packet size. We've seen huge SQL statements in production.

tests/test_reporter.py
yurishkuro: the test was unstable when it relied on the difference between 1ms and 5ms

tests/test_span.py
yurishkuro: String truncation works fine with unicode. I don't see a point of testing a feature of the language.

```
>>> x=u'key2-caf\xe9-blah'
>>> print(x)
key2-café-blah
>>> print(x[:10])
key2-café-
>>> print(x[:9])
key2-café
>>> print(x[:8])
key2-caf
```

jaeger_client/constants.py
vprithvi: If we want to figure out the size in bytes, I think it is more useful to have `MAX_TAG_VALUE_BYTES` instead of length, because the length would be shorter than the number of bytes when using some unicode characters.

yurishkuro: why does it matter? It's an approximation anyway.

vprithvi: I see, I didn't realize that it was an approximation.

PR #111, opened 2017-12-18
Replace zipkin.thrift out-of-band span format with jaeger.thrift

### LINKED ISSUE #87: Update python client to use jaeger model instead of zipkin

yurishkuro:
Fixes #87


yurishkuro:
Currently extends #110, once #110 is merged I will rebase this off master

jaeger_client/thrift.py
```diff
@@ -12,17 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from past.builtins import basestring
 import six
 import socket
 import struct
 
-import jaeger_client.thrift_gen.zipkincore.ZipkinCollector as zipkin_collector
+import jaeger_client.thrift_gen.jaeger.ttypes as ttypes
```
black-adder: is the plan to add a zipkinthrift converter and a zipkin http reporter? If so, a lot of this code seems pretty useful to keep track of.

jaeger_client/thrift.py
```diff
@@ -121,59 +69,50 @@ def timestamp_micros(ts):
     return long(ts * 1000000)
 
 
-def make_zipkin_spans(spans):
-    zipkin_spans = []
+def make_tags(tags):
+    # TODO extend to support non-string tag values
+    return [
+        make_string_tag(key=k, value=str(v), max_length=1024)
```
black-adder: MAX_TAG_VALUE_LENGTH?

yurishkuro: well if that then better to use the configurable value, which will need to be pushed through. I can do it.

jaeger_client/thrift.py
yurishkuro: the plan is to drop zipkin format. It's not useful over UDP, but if someone wants to implement an http sender with zipkin.thrift, this code can be resurrected from git history. I don't want to waste time on keeping it since nobody asked for it.

PR #112, opened 2017-12-19
Add codec for B3 trace

### LINKED ISSUE #106: Zipkin HTTP headers support
Hello,

I'm currently working on adding Zipkin HTTP headers support to this library and would like to discuss what the preferred method for enabling support for those headers should be.

My current solution is to have a new config parameter, something like `use_zipkin_http_headers: True`, which would send the Zipkin codec as `extra_codecs` to the Tracer and take over handling of HTTP headers over the default TextCodec. 

Is this the recommended way of adding codecs to the tracer? If not, how should I be enabling the Zipkin support?

gravelg:
A new config parameter called `b3_codec` is available to use the B3 codec with HTTP headers

Fixes #106 

Signed-off-by: Gabriel Gravel <ggravel@appnexus.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,15 +133,13 @@ This allows using Jaeger UI to find the trace by this tag.
 
 This library internally uses Zipkin Thrift data model and conventions, 
 but if you want to use it directly with other Zipkin libraries & backend, 
-it needs:
-  1. different [wire codecs](./jaeger_client/codecs.py) to transmit 
-     trace context as `X-B3-*` headers
-  2. a reporter that will submit traces to Zipkin backend over Zipkin-supported 
-     transports like Kafka or HTTP
+you can provide the configuration property `propagation: 'b3'` and the
+`X-B3-*` HTTP headers will be supported.
 
-Both of these things are easy to add (e.g. it was done in https://github.com/jaegertracing/jaeger-client-java/pull/34), 
-but it is not a priority for the Uber team since we are using a different backend.
-We will welcome PRs that provide that functionality.
+The B3 codec assumes it will receive lowercase HTTP headers, as this seems
+to be the standard in the popular frameworks like Flask and Django.
+Please make sure your framework does the same.
+
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -131,8 +131,7 @@ This allows using Jaeger UI to find the trace by this tag.
 
 ## Zipkin Compatibility
 
-This library internally uses Zipkin Thrift data model and conventions, 
-but if you want to use it directly with other Zipkin libraries & backend, 
+To use this library directly with other Zipkin libraries & backend,
 you can provide the configuration property `propagation: 'b3'` and the
 `X-B3-*` HTTP headers will be supported.
 
--- a/jaeger_client/codecs.py
+++ b/jaeger_client/codecs.py
@@ -235,18 +235,16 @@ class ZipkinCodec(Codec):
                            baggage=None)
 
 
-def header_to_hex(header, trace_header=False):
+def header_to_hex(header):
     if not isinstance(header, (str, unicode)):
         raise SpanContextCorruptedException(
-            'malformed trace context "%s", expected 16 or 32 character hex string' % header)
-    if len(header) != 16 and (trace_header and len(header) != 32):
-        raise SpanContextCorruptedException(
-            'malformed trace context "%s", expected 16 or 32 character hex string' % header)
+            'malformed trace context "%s", expected hex string' % header)
+
     try:
         return int(header, 16)
     except ValueError:
         raise SpanContextCorruptedException(
-            'malformed trace context "%s", expected 16 or 32 character hex string' % header)
+            'malformed trace context "%s", expected hex string' % header)
 
 
 class B3Codec(Codec):
@@ -264,16 +262,16 @@ class B3Codec(Codec):
     def inject(self, span_context, carrier):
         if not isinstance(carrier, dict):
             raise InvalidCarrierException('carrier not a dictionary')
-        carrier[self.trace_header] = format(span_context.trace_id, 'x')
-        carrier[self.span_header] = format(span_context.span_id, 'x')
+        carrier[self.trace_header] = format(span_context.trace_id, 'x').zfill(16)
+        carrier[self.span_header] = format(span_context.span_id, 'x').zfill(16)
         if span_context.parent_id is not None:
-            carrier[self.parent_span_header] = format(span_context.parent_id, 'x')
+            carrier[self.parent_span_header] = format(span_context.parent_id, 'x').zfill(16)
         carrier[self.flags_header] = str(span_context.flags)
 
     def extract(self, carrier):
         if not isinstance(carrier, dict):
             raise InvalidCarrierException('carrier not a dictionary')
-        trace_id = header_to_hex(carrier.get(self.trace_header.lower()), trace_header=True)
+        trace_id = header_to_hex(carrier.get(self.trace_header.lower()))
         span_id = header_to_hex(carrier.get(self.span_header.lower()))
         parent_id = header_to_hex(carrier.get(self.parent_span_header.lower()))
         flags = 0x00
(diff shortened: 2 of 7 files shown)


jaeger_client/codecs.py
```diff
@@ -34,6 +34,9 @@
 import six
 import urllib.parse
 
+SAMPLED_FLAG = 0x01
+DEBUG_FLAG = 0x02
```
yurishkuro: it's better to import these from `.span`

jaeger_client/codecs.py
```diff
@@ -232,3 +235,50 @@ def extract(self, carrier):
         return SpanContext(trace_id=trace_id, span_id=span_id,
                            parent_id=parent_id, flags=flags,
                            baggage=None)
+
+
+def header_to_hex(header):
+    if not isinstance(header, (str, unicode)):
+        return None
+    return int(header, 16)
```
yurishkuro: note that invalid strings will cause ValueError, which breaks the contract of the codec. The main codec protects against that.

jaeger_client/codecs.py
```diff
@@ -232,3 +235,50 @@ def extract(self, carrier):
...
+        carrier[self.trace_header] = format(span_context.trace_id, 'x')
+        carrier[self.span_header] = format(span_context.span_id, 'x')
+        if span_context.parent_id is not None:
+            carrier[self.parent_span_header] = format(span_context.parent_id, 'x')
+        carrier[self.flags_header] = str(span_context.flags)
+
+    def extract(self, carrier):
+        if not isinstance(carrier, dict):
+            raise InvalidCarrierException('carrier not a dictionary')
+        trace_id = header_to_hex(carrier.get(self.trace_header.lower()))
```
yurishkuro: a. for better performance `lower()` could've been called in the constructor
b. if there actually a convention among python http frameworks to represent headers as a lower-case dict? Our main codec makes case-insensitive comparisons

jaeger_client/config.py
```diff
@@ -246,6 +248,14 @@ def tags(self):
                 tags[key.strip()] = value.strip()
         return tags
 
+    @property
+    def extra_codecs(self):
```
yurishkuro: This feels rather misleading, it's not adding "extra" codecs, it's overriding (partially) the existing ones.

I would much rather use the approach we took in https://github.com/jaegertracing/jaeger-client-java/pull/296 where we have just a single config property `codecs`, which is a comma-separated list like `jaeger,b3` or `jaeger,w3c`. We don't actually need to support the list right now, just single value `b3` that will replace the jaeger codecs.

jaeger_client/codecs.py
gravelg: I had tried that originally but `.span` already imports `codecs` so we end up with a circular import problem. Should I just move them to utils.py?

jaeger_client/config.py
gravelg: It seems the API is a little different for Tracer in the Java client and here, should I rename this parameter to codecs and add a new optional argument to the Tracer `__init__` method, or even rename the existing `extra_codecs` param that is already there? I was kinda just piggybacking on that being present to add the codec in a simple way

jaeger_client/codecs.py
gravelg: The popular Python http frameworks that I know all do `.lower()` on all http headers. I think that this is the most common approach for case insensitive comparisons (vs `.upper()`)

jaeger_client/codecs.py
yurishkuro: constants.py would be better

jaeger_client/config.py
yurishkuro: I think we don't need to change the Tracer API, and you can still pass the B3 codec as `extra_codecs` to it. I was mostly referring to the API of the Configuration class, it's there I would like to have `codecs=x,y,z` property.

jaeger_client/codecs.py
yurishkuro: ok, well we can start with this assumption. Perhaps add it to the readme. Which reminds me - please add/update a section in the readme documenting how the B3 codec can be configured.

gravelg:
I've updated the Readme, changed the configuration property to mimic the Java implementation and added some tests and validations for the codec.

README.md
```diff
@@ -133,15 +133,13 @@ This allows using Jaeger UI to find the trace by this tag.
 
 This library internally uses Zipkin Thrift data model and conventions,
```
yurishkuro: this is now out of date since I merged #111. Do you mind rephrasing that only in-band interop with Zipkin is possible with your change?

jaeger_client/codecs.py
```diff
@@ -232,3 +233,54 @@ def extract(self, carrier):
...
+def header_to_hex(header):
+    if not isinstance(header, (str, unicode)):
+        raise SpanContextCorruptedException(
+            'malformed trace context "%s", expected hex string' % header)
+    try:
+        cast_header = int(header, 16)
+    except ValueError:
+        raise SpanContextCorruptedException(
+            'malformed trace context "%s", expected hex string' % header)
+    return cast_header
```
yurishkuro: nit: can return directly from L243

jaeger_client/codecs.py
```diff
@@ -232,3 +233,54 @@ def extract(self, carrier):
...
+        self.trace_header = 'X-B3-TraceId'
+        self.span_header = 'X-B3-SpanId'
+        self.parent_span_header = 'X-B3-ParentSpanId'
+        self.sampled_header = 'X-B3-Sampled'
+        self.flags_header = 'X-B3-Flags'
+
+    def inject(self, span_context, carrier):
+        if not isinstance(carrier, dict):
+            raise InvalidCarrierException('carrier not a dictionary')
+        carrier[self.trace_header] = format(span_context.trace_id, 'x')
```
yurishkuro: are you sure `format(n, 'x')` is the right format for Zipkin?  I seem to recall it being rather pendantic about the exact length of the headers ["32 or 16 lower-hex characters"](https://github.com/openzipkin/b3-propagation#traceid-1).

gravelg: I'll add a size validation for those things

yurishkuro: hi, sorry if I wasn't clear - I was referring to the `inject` method. The `format` function will not do 0-padding, so you are not guaranteed to produce 16/32 hex strings.

In the reading side (header_to_hex) I would actually be in favor of allowing any length, just like Jaeger does.

gravelg: Totally my mistake, I got too focused the hex format on the openzipkin spec. I'll do the correct change

gravelg: Also, it seems the java refuses hex strings of the wrong length https://github.com/jaegertracing/jaeger-client-java/blob/17b7a2c448a2d6beacede6e07e134580a2250493/jaeger-core/src/main/java/com/uber/jaeger/propagation/HexCodec.java

Should I keep or remove the validation?

yurishkuro: it refuses strings >32 as far as I can tell, but not anything in between.

yurishkuro:
PS - please rebase from master

tests/test_codecs.py
```diff
@@ -299,6 +299,60 @@ def test_zipkin_codec_inject(self):
...
+
+        with self.assertRaises(InvalidCarrierException):
+            codec.inject(span_context=None, carrier=[])
+
+        ctx = SpanContext(trace_id=256, span_id=127, parent_id=None, flags=1)
+        span = Span(context=ctx, operation_name='x', tracer=None, start_time=1)
+        carrier = {}
+        codec.inject(span_context=span, carrier=carrier)
+        assert carrier == {'X-B3-SpanId': format(127, 'x').zfill(16),
+                           'X-B3-TraceId': format(256, 'x').zfill(16), 'X-B3-Flags': '1'}
```
yurishkuro: > `format(127, 'x').zfill(16)`

nit: this is a faulty way of writing tests, because it pretty much boils down to "to test my production function f(x) I am going to assert f(x) == f(x)". A better test is to compare the result with a value that is not produced by the same code as prod code, i.e. `'X-B3-SpanId': 0x000000000000007F`.

yurishkuro:
Thanks for contributing this, @gravelg !

PR #113, opened 2018-01-13
python3 compat, hasattr iteritems->items

kbroughton:
I'm not sure what data structures this test was for, but for python3 it fails because dict has no iteritems.  Changes were made around it for six.iteritems, but this change was missed.

Once i made this change (and another PR to django_opentracing), i was able to run django_opentracing on django python3

### DIFF
--- a/jaeger_client/codecs.py
+++ b/jaeger_client/codecs.py
@@ -82,7 +82,7 @@ class TextCodec(Codec):
                 carrier[header_key] = encoded_value
 
     def extract(self, carrier):
-        if not hasattr(carrier, 'iteritems'):
+        if not hasattr(carrier, 'items'):
             raise InvalidCarrierException('carrier not a collection')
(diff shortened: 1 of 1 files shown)


jaeger_client/codecs.py
```diff
@@ -82,7 +82,7 @@ def inject(self, span_context, carrier):
                 carrier[header_key] = encoded_value
 
     def extract(self, carrier):
-        if not hasattr(carrier, 'iteritems'):
+        if not hasattr(carrier, 'items'):
```
yurishkuro: I am not able to find the reason why this code was checking for `iteritems` instead of `isinstance(carrier, dict)` as done in the inject method, but I seem to recall there was a reason for it, i.e. a type was passed in some cases that wasn't a true dict but allowed iteritems (with there was a unit test).

I would feel safer if this check was backwards compatible, i.e. check for `items` OR `iteritems`.

PR #117, opened 2018-01-15
Fix span and sampler tests to work under Py3

yurishkuro:
Related to #59


yurishkuro:
cc @black-adder

PR #118, opened 2018-01-16
Remove unused import at README

eundoosong:
Remove "import opentracing" because we don't use.


yurishkuro:
thanks @eundoosong

PR #119, opened 2018-01-17
Resolve tag unicode issues

ekelleyv:
Context:
Getting errors when running [`before_request`](https://github.com/uber-common/opentracing-python-instrumentation/blob/master/opentracing_instrumentation/http_server.py#L35) on a tornado request that has unicode characters in the URL. 

From digging into tornado source, it looks like here they convert the url to `latin1`:
https://github.com/tornadoweb/tornado/blob/023a9ac87aa2de87585b71b8fee213346f4b590c/tornado/wsgi.py#L193-L194

https://github.com/tornadoweb/tornado/blob/023a9ac87aa2de87585b71b8fee213346f4b590c/tornado/wsgi.py#L55-L62

So the `http.url` being set as a tag in the opentracing instrumentation is actually a `latin1` encoding of those unicode characters. Then, when we try and cast it to the future `str`, It tries to decode again it appears, resulting in:
`UnicodeDecodeError: 'ascii' codec can't decode byte 0xe9 in position 3: ordinal not in range(128)`

I have recreated essentially what tornado is doing in the test cases here and put together a really quick solution. We would almost certainly want to factor this string handling out of `set_tag` but I will let someone more familiar with this codebase give me guidance on where that should go.

### DIFF
--- a/jaeger_client/span.py
+++ b/jaeger_client/span.py
@@ -86,15 +86,12 @@ class Span(opentracing.Span):
                     self.context.flags &= ~SAMPLED_FLAG
             elif self.is_sampled():
 
-                if isinstance(value, basestring):
-                    try:
-                        value = value.decode('latin1')
-                    except UnicodeEncodeError:
-                        pass
+                if not isinstance(value, basestring):
+                    value = str(value)
 
                 tag = thrift.make_string_tag(
                     key=key,
-                    value=str(value),
+                    value=value,
                     max_length=self.tracer.max_tag_value_length,
                 )
                 self.tags.append(tag)
--- a/jaeger_client/span.py
+++ b/jaeger_client/span.py
@@ -85,6 +85,13 @@ class Span(opentracing.Span):
                 else:
                     self.context.flags &= ~SAMPLED_FLAG
             elif self.is_sampled():
+
+                if isinstance(value, basestring):
+                    try:
+                        value = value.decode('latin1')
+                    except UnicodeEncodeError:
+                        pass
+
                 tag = thrift.make_string_tag(
                     key=key,
                     value=str(value),
--- a/tests/test_span.py
+++ b/tests/test_span.py
@@ -224,3 +224,12 @@ def test_span_tag_value_max_length(tracer):
     tag_n = len(span.tags) - 1
     assert span.tags[tag_n].key == 'x'
     assert span.tags[tag_n].vStr == 'x' * 42
+
+
+def test_span_tag_value_unicode(tracer):
+    span = tracer.start_span(operation_name='x')
+    latin1_encoded_string = u'caf\xe9'.encode('latin1')
+    span.set_tag('x', latin1_encoded_string)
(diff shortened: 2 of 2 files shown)


yurishkuro:
is it possible to include the complete stack trace leading to UnicodeDecodeError? Or because it's tornado the stack is meaningless?

ekelleyv:
Here is the relevant part:
```python
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 95: ordinal not in range(128)
 ...
  span = http_server.before_request(request=request_wrapper)
  File "opentracing_instrumentation/http_server.py", line 84, in before_request
    tags=tags_dict)
  File "jaeger_client/tracer.py", line 162, in start_span
    tags=tags, start_time=start_time)
  File "jaeger_client/span.py", line 54, in __init__
    self.set_tag(k, v)
  File "jaeger_client/span.py", line 96, in set_tag
    tag = thrift.make_string_tag(key, str(value))
  File "future/types/newstr.py", line 102, in __new__
    return super(newstr, cls).__new__(cls, value)
```

ekelleyv:
If instead you decide that this is tornado-specific, I could see an argument being made for doing this here rather than the client itself:
https://github.com/uber-common/opentracing-python-instrumentation/blob/master/opentracing_instrumentation/http_server.py#L140-L142

ekelleyv:
Ya the more I think about it, the more I think this kinda needs to be solved at a higher level. There is no way to know what encoding the string uses, `latin1` works in this case because we know its coming from tornado, but that is not a generic answer here.

I would suggest we make `TornadoRequestWrapper.full_url` (prob both in server and tornado_http) in `opentracing-instrumentation` return unicode rather than the `latin1` encoded string.

@black-adder I know you suggested solving it in jaeger client. What do you think?

yurishkuro:
yes, the use of latin1 in jaeger-client certainly doesn't look right. What's not clear to me is if we already have a string object why do we need to re-string it?

ekelleyv:
You suggesting something like?:
```python
if not isinstance(value, basestring):
    value = str(value)

tag = thrift.make_string_tag(
    key=key,
    value=value,
    max_length=self.tracer.max_tag_value_length,
)
```

ekelleyv:
I think that update should at least handle this issue gracefully (assuming things handle the latin1 string down the line properly).

ekelleyv:
@yurishkuro Lemme work on confirming that today. Will ship pointing at this branch and ensure that our sentries go away.

yurishkuro:
@ekelleyv any update?

ekelleyv:
Unfortunately the memory leak introduced when using jaeger + new relic together has mean I need to shelve our jaeger work right now. I did not get a chance to actually run this in production. Hopefully I can investigate the memory issues next week, but will need to do that before deploying jaeger to prod again.

yurishkuro:
@ekelleyv any update?

yurishkuro:
closing for inactivity

PR #121, opened 2018-02-05
Tidying up makefile

eundoosong:
1. Change zipkinCore to zipkincore(all lower-case) and remove unnecessary mv command
: Since the name of thrift file is zipkincore.thirft, we don't need to keep zipkinCore.thrift and mv cmd.
2. Change jaeger_client/thrift_gen to ${THRIFT_GEN_DIR}
: Just for consistency


eundoosong:
Hello, I think this is not source code change,  how coverage decreased?

PR #126, opened 2018-02-17
`urllib.parse.quote` returns future...newstr that causes bytes(str(u'foober'), 'utf8') == b"b'foober'"

dray92:
Code comments and following open issues in future library will explain the core problem.

[#171](https://github.com/PythonCharmers/python-future/issues/171), [#193](https://github.com/PythonCharmers/python-future/issues/193)

Signed-off-by: Debosmit Ray <debo@uber.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,12 @@ matrix:
     env: COVER=1
   - python: '2.7'
     env: CROSSDOCK=1
+  - python: '2.7.11'
+    env: CROSSDOCK=1
+  - python: '2.7.12'
+    env: CROSSDOCK=1
+  - python: '2.7.13'
+    env: CROSSDOCK=1
   - python: '3.3'
     env: CROSSDOCK=1
   - python: '3.4'
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,14 @@ matrix:
     env: COVER=1
   - python: '2.7'
     env: CROSSDOCK=1
+  - python: '3.3'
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -10,6 +10,22 @@ matrix:
...
+    env: CROSSDOCK=1
+  - python: '2.7.13'
+    env: CROSSDOCK=1
+  - python: '3.3'
+    env: CROSSDOCK=1
+  - python: '3.4'
+    env: CROSSDOCK=1
+  - python: '3.5'
+    env: CROSSDOCK=1
+  - python: '3.6'
```
yurishkuro: Jaeger currently doesn't work with 3.x, what's the reason for adding it?

Also, what's the reason for adding so many 2.7 minor versions, is there really any difference? It will make the tests run forever.

dray92: Ah - I was under the impression that we were supporting 3.x as well. Will remove.

All these 2.7.x version are currently in use at Uber. My reasoning for this was that the tests in this repo does not seems to take very long, so it is better to prevent future surprises.

dray92:
Something like 
```
>>> from future.types.newstr import newstr
>>> a = newstr('foo')
>>> type(a)
<class 'future.types.newstr.newstr'>
>>> b = a.encode('utf-8')
>>> type(b)
<class 'future.types.newbytes.newbytes'>
>>> b
b'foo'
>>> b.__repr__()
"b'foo'"
```
will not be solved by this...

PR #127, opened 2018-02-17
set_baggage_item on non-ascii key-value pairs would throw error

dray92:
Signed-off-by: Debosmit Ray <debo@uber.com>

### DIFF
--- a/jaeger_client/span.py
+++ b/jaeger_client/span.py
@@ -114,8 +114,8 @@ class Span(opentracing.Span):
         if self.is_sampled():
             logs = {
                 'event': 'baggage',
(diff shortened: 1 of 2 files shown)


dray92:
@yurishkuro tbh not sure if this behavior should be encouraged

yurishkuro:
@dray92 I think we don't want this, correct?

dray92:
@yurishkuro It depends. At the moment, present of non-ascii baggage item blows up in thrift rw serialization layer (crossdock tests should show that).
I don't know if this is a fight we want to fight now tbh. I'll let you decide.

PR #130, opened 2018-02-27
Use only `six` for py2/py3 compatibility, drop `future`

yurishkuro:
After installing `pycurl>=7.43,<8` and configuring tornado async client to use curl client in `testconf.py`:
```python
AsyncHTTPClient.configure('tornado.curl_httpclient.CurlAsyncHTTPClient')
```

some tests start failing with:
```
___________________ test_trace_propagation[HTTP-HTTP-True]
tests/test_crossdock.py:115: in test_trace_propagation
    assert tr.downstream.span.baggage == level1.get("baggage")
E   assert "b'Zoidberg'" == 'Zoidberg'
E     - b'Zoidberg'
E     ? --        -
E     + Zoidberg
```

Special thanks to @dray92 for narrowing it down.

The reason is because the `future` module implements `urllib.parse.quote` to return a `newstr` object that in turn returns `newbytes` object after encoding, and the latter gets serialized with the `b` prefix, causing the breaks above.

The fix is to stop using `future` module (it was only needed in one file) and instead use `six.moves.urllib.parse`, which fixes the tests.

### DIFF
--- a/jaeger_client/codecs.py
+++ b/jaeger_client/codecs.py
@@ -14,8 +14,8 @@
 
 from __future__ import absolute_import
 
-from future import standard_library
-standard_library.install_aliases()  # noqa
+# from future import standard_library
+# standard_library.install_aliases()  # noqa
 
 from builtins import object
 from past.builtins import basestring
@@ -33,7 +33,7 @@ from .span_context import SpanContext
 from .constants import SAMPLED_FLAG, DEBUG_FLAG
 
 import six
-import urllib.parse
+from six.moves import urllib_parse
 
 
 class Codec(object):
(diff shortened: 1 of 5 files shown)


dray92:
@yurishkuro looks awesome

It would be awesome to get that test in though, to prevent against regression.

yurishkuro:
@dray92 I couldn't find the test in your branch, the one with `b'value'` and tornado's async client

isaachier:
I have a test case that I believe would prove this is the correct fix. I added it after `tracer.inject` in `test_codecs.py`, test case `test_non_ascii_baggage_with_httplib`:

```
+        if isinstance(b[0], six.PY2.text_type):
+            key = '%s%s' % (BAGGAGE_HEADER_PREFIX, b[0].encode('utf8'))
+        else:
+            key = '%s%s' % (BAGGAGE_HEADER_PREFIX, b[0])
+        assert key in headers
```

jaeger_client/codecs.py
```diff
@@ -14,9 +14,6 @@
 
 from __future__ import absolute_import
 
-from future import standard_library
-standard_library.install_aliases()  # noqa
-
 from builtins import object
 from past.builtins import basestring
```
sc68cal: You dropped the `futures` package from the requirements but are still referencing it.

I get a stacktrace when trying to run this in a Python 3 environment.

```
  File "/usr/src/app/src/jaeger-client/jaeger_client/tracer.py", line 30, in <module>
  File "/usr/src/app/src/jaeger-client/jaeger_client/codecs.py", line 18, in <module>
    from past.builtins import basestring
ModuleNotFoundError: No module named 'past'
```

yurishkuro: Thanks for reporting - https://github.com/jaegertracing/jaeger-client-python/issues/133

PR #134, opened 2018-02-28
Replace `past` with `six`

### LINKED ISSUE #133: We dropped dependency on `future` but still have references to `past`
@sc68cal [said](https://github.com/jaegertracing/jaeger-client-python/pull/130#discussion_r171393944):

You dropped the future package from the requirements but are still referencing it.

I get a stacktrace when trying to run this in a Python 3 environment.

```
File "/usr/src/app/src/jaeger-client/jaeger_client/tracer.py", line 30, in <module>
  File "/usr/src/app/src/jaeger-client/jaeger_client/codecs.py", line 18, in <module>
    from past.builtins import basestring
ModuleNotFoundError: No module named 'past'
```

yurishkuro:
Fixes #133

cc @sc68cal @dray92

### DIFF
--- a/jaeger_client/thrift_gen/zipkincore/ttypes.py
+++ b/jaeger_client/thrift_gen/zipkincore/ttypes.py
@@ -5,7 +5,7 @@
 #
 #  options string: py:new_style,tornado
 #
-from past.builtins import xrange
+from six.moves import xrange
 
 from thrift.Thrift import TType, TMessageType, TException, TApplicationException
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ It can be fixed with:
 
 ```
 pip uninstall pycurl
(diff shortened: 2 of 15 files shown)


jaeger_client/thrift_gen/zipkincore/ttypes.py
```diff
@@ -5,7 +5,7 @@
 #
 #  options string: py:new_style,tornado
 #
-from past.builtins import xrange
+from six.moves import xrange
```
black-adder: if we generate thrift again, will we run into the same problem again?

yurishkuro: no, I fixed `thrift-gen-fix.awk`

yurishkuro:
The py2 env still contains both `future` and `past`, most likely pulled transitively by other dependencies. It's possible that in py3 these deps are not getting pulled. But we don't test with py3 yet (some test are still failing)l

PR #129, opened 2018-02-24
Make Metrics consistent with Go client.

eundoosong:
Make Metrics consistent with Go client.

Now the metric name shows consistent with go-client as you see below prometheus test I did.
- in sampler, add SamplerMetrics which has 4 types of metrics, as go-client has.
- in tracer, add spans_finished 

Resolves #102 

Please review.

```
# HELP jaeger:traces jaeger:traces
# TYPE jaeger:traces counter
jaeger:traces{sampled="n",state="started"} 6.0
jaeger:traces{sampled="y",state="started"} 0.0
jaeger:traces{sampled="y",state="joined"} 0.0
jaeger:traces{sampled="n",state="joined"} 0.0
# HELP jaeger:sampler_updates jaeger:sampler_updates
# TYPE jaeger:sampler_updates counter
jaeger:sampler_updates{result="err"} 0.0
jaeger:sampler_updates{result="ok"} 0.0
# HELP python_info Python platform information
# TYPE python_info gauge
python_info{implementation="CPython",major="2",minor="7",patchlevel="10",version="2.7.10"} 1.0
# HELP jaeger:sampler_queries jaeger:sampler_queries
# TYPE jaeger:sampler_queries counter
jaeger:sampler_queries{result="err"} 0.0
jaeger:sampler_queries{result="ok"} 0.0
# HELP jaeger:started_spans jaeger:started_spans
# TYPE jaeger:started_spans counter
jaeger:started_spans{sampled="n"} 12.0
jaeger:started_spans{sampled="y"} 0.0
# HELP jaeger:finished_spans jaeger:finished_spans
# TYPE jaeger:finished_spans counter
jaeger:finished_spans 0.0
# HELP jaeger:reporter_spans jaeger:reporter_spans
# TYPE jaeger:reporter_spans counter
jaeger:reporter_spans{result="dropped"} 0.0
jaeger:reporter_spans{result="socket_error"} 0.0
jaeger:reporter_spans{result="err"} 0.0
jaeger:reporter_spans{result="ok"} 0.0
```

Signed-off-by: eundoo-song <eundoo.song@gmail.com>

PR #140, opened 2018-03-16
Add PrometheusMetricsFactory

eundoosong:
As discussed in [PR 129](https://github.com/jaegertracing/jaeger-client-python/pull/129) with @yurishkuro, PrometheusMetricsFactory is ready for review.

### DIFF
--- a/jaeger_client/metrics_factory/README.md
+++ b/jaeger_client/metrics_factory/README.md
@@ -1,0 +1,22 @@
+# Prometheus integration for Jaeger internal metrics
+
+This module brings a [Prometheus](https://github.com/prometheus/client_python) integration to the internal Jaeger metrics.
+
+## Installation
+
+```
+pip install prometheus_client
+```
+
+## Configuring
+
+To configure the Prometheus integration, a custom Jaeger Tracer instance has to be used:
+
+```python
+config = Config(
+        service_name='your-app-name',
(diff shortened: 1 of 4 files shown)


eundoosong:
I'll fix DCO.

@yurishkuro 
If you see the coverage result, it is decreased because pytest-cov works as if there is no unit-testing for prometheus_metrics.py
I don’t want to run unit-testing(pytest) and coverage(pytest-cov) if prometheus_client is not installed.
skipif decorator provided by pytest is added in test_prometheus_metrics.py for unit-testing skipping, which works good.  
```python 
@pytest.mark.skipif('prometheus_client' not in sys.modules,  
                                     reason="requires prometheus_client")
class TestPrometheusMetricsFactory:
    def test_prometheus_metrics_counter(self):
        ....
```
But I can’t find any similar way(conditional skip) to do the same in pytest-cov.
Options only provided by pytest-cov is adding "exclude list" into .coveragerc’s omit for file OR "# pragma: no cover" for code in a file.
They are fixed list, not like conditional thing,  and no co-working with pytest

I have no idea how get the coverage passed in ci-testing. 
Would you please give better suggestion on this?

Alternatively, I prepared update-coveragerc.sh for adding a file path in “omit” when a library is not installed. If it looks good, I can merge this.
https://github.com/eundoosong/jaeger-client-python/blob/prometheus_poc/scripts/update-coveragerc.sh

Issue #139, opened 2018-03-16
Include installation for Windows and MacOS in README


PikBot:
Currently, the installation instructions are only for LINUX. 

Separate installation instruction would be desirable for MacOS and Windows users.

yurishkuro:
I don't think we even need installation instructions

PikBot:
In that case, is there a need to remove those installation instructions? Especially the one specific to LINUX : `apt-get install python-dev`because it could be quite misleading for MacOS and Windows users.

yurishkuro:
yes, I would just leave the `pip install` line

PR #142, opened 2018-03-16
Add PrometheusMetricsFactory

eundoosong:
As discussed in [PR 129](https://github.com/jaegertracing/jaeger-client-python/pull/129) with @yurishkuro, PrometheusMetricsFactory is ready for review.


eundoosong:
@yurishkuro 
If you see the coverage result, it is decreased because pytest-cov works as if there is no unit-testing for prometheus_metrics.py
I don’t want to run unit-testing(pytest) and coverage(pytest-cov) on PrometheusMetricsFactory if prometheus_client is not installed.
skipif decorator provided by pytest is added in test_prometheus_metrics.py for unit-testing skipping, which works good.  
```python 
@pytest.mark.skipif('prometheus_client' not in sys.modules,  
                                     reason="requires prometheus_client")
class TestPrometheusMetricsFactory:
    def test_prometheus_metrics_counter(self):
        ....
```
But I can’t find any similar way(conditional skip) to do the same in pytest-cov.
Options only provided by pytest-cov is adding "exclude list" into .coveragerc’s omit for file OR "# pragma: no cover" for code in a file.
They are fixed list, not like conditional thing,  and no co-working with pytest

I have no idea how get the coverage passed in ci-testing. 
Would you please give better suggestion on this?

Alternatively, I prepared update-coveragerc.sh for adding a file path in “omit” when a library is not installed. If it looks good, I can merge this.
https://github.com/eundoosong/jaeger-client-python/blob/prometheus_poc/scripts/update-coveragerc.sh

yurishkuro:
> I don’t want to run unit-testing(pytest) and coverage(pytest-cov) on PrometheusMetricsFactory if prometheus_client is not installed.

Why not always run them?  prometheus_client can be declared as dev dependency, as we do with other modules like opentracing instrumentation or tchannel (https://github.com/jaegertracing/jaeger-client-python/blob/master/setup.py#L64)

README.md
```diff
@@ -14,7 +14,6 @@ Please see [CONTRIBUTING.md](./CONTRIBUTING.md).
 ## Installation
 
 ```bash
-apt-get install python-dev
```
eundoosong: This is already removed by https://github.com/jaegertracing/jaeger-client-python/issues/139
Not related to this. ( committed  by mistake )

eundoosong:
@yurishkuro 
1. prometheus_client  as dev dependency 
Agree,  it is more simple and clear. Thanks. 

1. What benefits do you see in placing the new class in a subdirectory instead of, say, jaeger_client/prometheus.py?
I think prometheus_metrics is somewhat not part of jaeger client, it's more like an optional plugin for metrics purpose.
And there might be other metrics coming in the future too.  So, I moved it to the subdirectory. 
Please tell me your thought,

1. Add Prometheus instructions in main readme.
Agree.

README.md
```diff
@@ -138,6 +138,21 @@ The B3 codec assumes it will receive lowercase HTTP headers, as this seems
 to be the standard in the popular frameworks like Flask and Django.
 Please make sure your framework does the same.
 
+## Prometheus metrics
```
yurishkuro: I think it's better to move this to L83 as level-4 header, i.e. as a subsection of Configuration/Production.

README.md
```diff
@@ -138,6 +138,21 @@ The B3 codec assumes it will receive lowercase HTTP headers, as this seems
...
+This module brings a [Prometheus](https://github.com/prometheus/client_python) integration to the internal Jaeger metrics.  
+The way to initialize the tracer with Prometheus metrics:  
+
+```python
+from jaeger_client.metrics_factory.prometheus_metrics import PrometheusMetricsFactory
+
+config = Config(
+        config={},
+        service_name='your-app-name',
+        metrics_factory=PrometheusMetricsFactory(namespace=service_name)
```
yurishkuro: maybe `namespace='your-app-name'` for consistency?

eundoosong:
>instead of this
>from jaeger_client.**metrics_factory**.prometheus_metrics import PrometheusMetricsFactory
>
>support this
>from jaeger_client.**metrics**.prometheus import PrometheusMetricsFactory
>

Earlier, I named metrics as subdirectory, but naming conflict came like below since **metrics**.py and **metrics** subdirectory exists in the same location.
So I simply renamed to metrics_factory to resolve the conflict.
If you have better idea (or better naming), please tell me.

```shell
$ python example_prometheus.py
Traceback (most recent call last):
  File "example_prometheus.py", line 3, in <module>
    from jaeger_client import Config
  File "/Users/eundusong/dev/opensource/jaeger/dev/jaeger-client-python/jaeger_client/__init__.py", line 26, in <module>
    from .tracer import Tracer  # noqa
  File "/Users/eundusong/dev/opensource/jaeger/dev/jaeger-client-python/jaeger_client/tracer.py", line 33, in <module>
    from .metrics import Metrics, LegacyMetricsFactory
ImportError: cannot import name Metrics
```

yurishkuro:
> Earlier, I named metrics as subdirectory, but naming conflict error came like below since metrics.py ...

I see. I just put a PR #143 that moves metrics.py into a subdirectory. It looks like it's backwards compatible. Let's merge that first, then you can rebase and move your code to `jaeger_client/metrics/prometheus.py`

eundoosong:
@yurishkuro 
moved to "metrics/"
Ready to be merged, or is there anything else missing?

README.md
eundoosong: moved

README.md
eundoosong: add namespace='your-app-name'

yurishkuro:
looks great. Thanks!

PR #146, opened 2018-03-29
Change itermitems to Support Python 3

shblhy:
* change dict.iteritems() to six.iteritems(dict)
* change _to_string return str not bytes

Signed-off-by: Yan Huang <345471536@qq.com>

### DIFF
--- a/jaeger_client/thrift.py
+++ b/jaeger_client/thrift.py
@@ -51,6 +51,8 @@ def id_to_int(big_id):
 
 def _to_string(s):
     try:
+        if isinstance(s, str):
+            return s
         if isinstance(s, six.text_type):  # This is unicode() in Python 2 and str in Python 3.
             return s.encode('utf-8')
         else:
--- a/jaeger_client/thrift_gen/agent/Agent.py
+++ b/jaeger_client/thrift_gen/agent/Agent.py
@@ -209,7 +209,7 @@ class emitZipkinBatch_args(object):
 
   def __repr__(self):
     L = ['%s=%r' % (key, value)
-      for key, value in self.__dict__.iteritems()]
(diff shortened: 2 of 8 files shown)


jaeger_client/thrift.py
```diff
@@ -51,6 +51,8 @@ def id_to_int(big_id):
 
 def _to_string(s):
     try:
+        if isinstance(s, str):
+            return s
```
yurishkuro: this will skip encoding in py3

jaeger_client/thrift_gen/agent/Agent.py
```diff
@@ -209,7 +209,7 @@ def __hash__(self):
 
   def __repr__(self):
     L = ['%s=%r' % (key, value)
-      for key, value in self.__dict__.iteritems()]
+      for key, value in six.iteritems(self.__dict__)]
```
yurishkuro: this makes sense, but we don't edit generated files manually, there is a section in the makefile that rewrites some of the code.

shblhy: I think you use thrift and .thirft file to gen ttypes.py file,  where can i get the .thirft files?

yurishkuro: `.thrift` files are in `jaeger-idl`, run `git submodule update --init`

yurishkuro: actually `make thrift` will do everything.

yurishkuro: see https://github.com/jaegertracing/jaeger-client-python/blob/master/CONTRIBUTING.md#thrift

yurishkuro:
Py3 compatibility has been achieved with other PRs

Issue #149, opened 2018-04-11
Better support for multiple tracers/services


nziebart:
In some cases it is useful to simulate the existence of different services from within the same process, e.g. when making queries to services that are not instrumented (e.g. databases). In this case we'd want the `service` on the span to be the remote service.

However, the only way to do this is to create multiple tracers. I came across this proposal https://github.com/opentracing/specification/issues/77 for adding a `service` tag, but it has not been fleshed out.

The go client makes it pretty straightforward to create multiple tracers, by having methods on the `config` object to create a `Sampler` and a `Reporter`. However the python client does all the heavy lifting inside the `initialize_tracer` method, which can only be called once.

Should we break out factory methods here like we do in the go client? (see https://github.com/jaegertracing/jaeger-client-go/blob/master/config/config.go#L269).

yurishkuro:
> However the python client does all the heavy lifting inside the initialize_tracer method, which can only be called once.

do you actually need any heavy lifting?  Multiple tracers per process is still very much an edge case, I think it's reasonable to expect the application to do a bit of work to create them. I don't think it requires any new factory methods because the configuration class is merely a helper, it uses public constructors for the rest of the objects.

nziebart:
You're right, it is not strictly necessary. However in practice people will probably just copy the code from the config class if they want to do this, because there are certain bits of logic that are not really trivial - e.g. wrapping the reporter with a `CompositeReporter` and `LoggingReporter` if logging is enabled. It is a strange (and brittle) pattern to inspect the config in your own code and apply this logic - it feels better done by the library.

A simple proposal would be to just add methods on the `Config` class to create a `Reporter` and `Sampler`.

You mention that this is very much an edge case - is it common for users to have things like redis, kafka, mysql instrumented? It seems like users tend to wrap the client only and not the service (e.g. in https://github.com/opentracing-contrib/python-redis)

yurishkuro:
Fair enough. What specific methods would you like to propose? It looks like all you really need is `initialize_tracer`, maybe with an option to skip installing the global tracer.

An even better approach would be to actually implement support for setting service name via tag (per https://github.com/opentracing/specification/issues/77) and allow submitting spans for different services from the same tracer.

dmazin:
@nziebart Hi! This issue is from 4 years ago, but I just wanted to thank you for 1) identifying a common issue 2) actually following up with a PR. And thank you, maintainers, for being receptive and for merging the PR :-D

PR #150, opened 2018-04-11
Better support for creating multiple tracers

nziebart:
Make it easier to create multiple tracers, for creating spans around calls to services that are not themselves instrumented (e.g. databases).

See https://github.com/jaegertracing/jaeger-client-python/issues/149

### DIFF
--- a/jaeger_client/config.py
+++ b/jaeger_client/config.py
@@ -296,6 +296,16 @@ class Config(object):
                 return
             Config._initialized = True
 
+        tracer = self.new_tracer(io_loop)
+
+        self._initialize_global_tracer(tracer=tracer)
+        return tracer
+
+    def new_tracer(self, io_loop=None):
+        """
+        Create a new Jaeger Tracer based on the passed `jaeger_client.Config`.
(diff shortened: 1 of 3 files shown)


jaeger_client/config.py
```diff
@@ -296,6 +296,16 @@ def initialize_tracer(self, io_loop=None):
...
+
+        self._initialize_global_tracer(tracer=tracer)
+        return tracer
+
+    def new_tracer(self, io_loop=None):
+        """
+        Create a new Jaeger Tracer based on the passed `jaeger_client.Config`.
+        Does not set `opentracing.tracer` global variable.
+        """
         channel = self._create_local_agent_channel(io_loop=io_loop)
```
nziebart: Only downside here is that we create a separate channel for each tracer, when they could(?) use the same one. However, since the client can pass in an io loop it's not a trivial fix. Not sure how much of an issue this is.

yurishkuro:
not sure if the tests need to be adjusted, the coverage went down

nziebart:
There weren't any tests around `initialize_tracer`, so I just added a very basic one

Issue #22, opened 2016-10-26
Test handling of tags with Unicode characters


yurishkuro:
What happens during thrift serialization if tags contain Unicode characters?

Cf. https://github.com/lightstep/lightstep-tracer-python/pull/21

PR #154, opened 2018-04-16
Fix tests to work with Py 3.6

### LINKED ISSUE #59: Support Python 3.x

### LINKED ISSUE #137: 3.8.0 doesn't report spans in Python 3
I have a set of apps using the Go and Python jaeger client libs and the jaeger all-in-one docker image.  They're all running in a docker-compose stack - so the apps can see the jaeger container & reach it via its hostname 'jaeger'.  I also have the jaeger ports exposed on localhost so I can run sanity checks on the python code locally (e.g. jaeger all-in-one's docker ps looks like 0.0.0.0:5775->5775/udp, 5778/tcp, 0.0.0.0:6831-6832->6831-6832/udp, 14268/tcp, 0.0.0.0:16686->16686/tcp)

With 3.7.1 I was able to report spans correctly - running the python code locally while relying on the default reporter config (localhost:5775 in 3.7.1) and in docker (by setting local_agent/reporting_host to 'jaeger').

However, I had to upgrade to 3.8.0 to propagate spans from the Go app to Python app (because we're on Python 3 & 'extract' was relying on a Python 2 collection check.  In 3.8.0 the default reporter config is localhost:6831.

On 3.8.0 I'm unable to get any spans reported correctly from Python app, whether locally (relying on default config) or in docker (setting reporting_host to 'jaeger').  I'm also unable to get it to work changing the port back to 5775 (both locally or in docker).  

The logger still prints that it's reporting spans e.g. 
```
2018-03-08 12:23:23,226 Reporting span 346237e4f6ff27ab:346237e4f6ff27ab:0:1.....
```
but nothing makes it to the collector.

I see elsewhere that (#59) that Python3 is perhaps not supported - is that the root cause here? It seems like 3.8.0 was trying to address that somewhat.  Is there anything else I need to change when upgrading from 3.7.1 to 3.8.0?

### LINKED ISSUE #151: Thrift encoding breaks in python3
In python3, spans are unable to be serialized by thrift due to this line: https://github.com/jaegertracing/jaeger-client-python/blob/master/jaeger_client/thrift.py#L55

In python3, the `encode()` function returns a `bytes` literal, which then causes the following error in thrift, which is expecting a `str`:
```
...
  File "/Users/nziebart/.virtualenvs/myapp/lib/python3.6/site-packages/jaeger_client/thrift_gen/jaeger/ttypes.py", line 146, in write
    oprot.writeString(self.key)
  File "/Users/nziebart/.virtualenvs/myapp/lib/python3.6/site-packages/thrift/protocol/TProtocol.py", line 121, in writeString
    self.writeBinary(str_to_binary(str_val))
  File "/Users/nziebart/.virtualenvs/myapp/lib/python3.6/site-packages/thrift/compat.py", line 41, in str_to_binary
    return bytes(str_val, 'utf8')
TypeError: encoding without a string argument
```

It can be fixed by replacing the `_to_string` function linked above with:
```
        if six.PY2 and isinstance(s, unicode):
            return s.encode('utf8')
        else:
            return str(s)
```

I can open a PR for this tomorrow

yurishkuro:
* enable testing with Python 3.6
* some fixes to the codec to handle unicode strings with Py3
* a fix for thrift converter that broke Py3 compatibility in 3.8.0 release
* skip crossdock unit tests when on Py3
* remove tests for non-ascii characters in baggage keys - while those tests were passing with Py2, they didn't actually confirm that such baggage was readable on the receiving side, and it didn't really work. It's very hard to make it work because there's no way of knowing which encoding the strings are coming in from inbound request. This change **officially removes support for non-ascii characters in baggage keys.**

Fixes #151 
Fixes #137 
Fixes #59 
Partially solves #22 (missing roundtrip test with Unicode)

### DIFF
--- a/tests/test_codecs.py
+++ b/tests/test_codecs.py
@@ -118,6 +118,7 @@ class TestCodecs(unittest.TestCase):
                 'fry': u'Leela',
                 b'key1': bytes(chr(75)) if six.PY2 else bytes([75]),
                 u'key2': 'cafe',
+                u'key3': u'\U0001F47E',
             }
             carrier = {}
             codec.inject(ctx, carrier)
@@ -131,6 +132,7 @@ class TestCodecs(unittest.TestCase):
                     'trace-attr-fry': 'Leela',
                     'trace-attr-key1': 'K',
                     'trace-attr-key2': 'cafe',
+                    'trace-attr-key3': '%F0%9F%91%BE',
                 }, 'with url_encoding = %s' % url_encoding
                 for key, val in six.iteritems(carrier):
                     assert isinstance(key, str)
@@ -142,6 +144,7 @@ class TestCodecs(unittest.TestCase):
                     'trace-attr-fry': 'Leela',
                     'trace-attr-key1': 'K',
                     u'trace-attr-key2': 'cafe',
+                    'trace-attr-key3': u'\U0001F47E',
                 }, 'with url_encoding = %s' % url_encoding
 
     def test_context_from_bad_readable_headers(self):
@@ -416,6 +419,7 @@ def test_baggage_as_unicode_strings_with_httplib(httpserver):
         (u'key2', b'value'),
         ('key3', u'value'),
         (b'key4', bytes(chr(255)) if six.PY2 else bytes([255])),
+        (u'key5', u'\U0001F47E')
     ]
     for b in baggage:
         span = tracer.start_span('test')
--- a/tests/test_codecs.py
+++ b/tests/test_codecs.py
@@ -35,9 +35,6 @@ from opentracing.propagation import (
 )
 
 
-byte255 = bytes(chr(255)) if six.PY2 else bytes([255])
-
-
 class TestCodecs(unittest.TestCase):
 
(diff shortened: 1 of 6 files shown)


tests/test_codecs.py
```diff
@@ -134,10 +130,9 @@ def test_context_to_readable_headers(self):
                     'trace-id': '100:7f:0:1',
                     'trace-attr-bender': 'Countess%20de%20la%20Roca',
                     'trace-attr-fry': 'Leela',
-                    'trace-attr-key1': '%FF',
-                    'trace-attr-key2-caf\xc3\xa9': 'caf%C3%A9',
-                    'trace-attr-key3': 'caf%C3%A9',
-                    'trace-attr-key4-caf\xc3\xa9': 'value',
```
black-adder: isn't this still a useful testcase?

yurishkuro: updated the PR description: This change **officially removes support for non-ascii characters in baggage keys.**

PR #155, opened 2018-04-17
Replace concurrent.futures.Future with tornado.concurrent.Future

yurishkuro:


yurishkuro:
@black-adder `concurrent.futures` is standard in Py3 and `futures` is a backport of that into Py2 (that's why we can't depend on it in Py3, that module doesn't even compile w/ Py3 - although it could've)

On the other hand, `tornado.concurrent.Future` is a separate implementation that exists both in py2 and py3 (it might be just an alias to the official Future in py3).

PR #156, opened 2018-04-18
absolute_import and mark span as error on exception

bohea:
*  If exception has occurred, mark span as error
*  Use absolute_import introduced in python3.x
*  allow debug_id_header


jaeger_client/span.py
```diff
@@ -182,3 +187,17 @@ def error(self, message, payload=None):
         else:
             self.log(event=message)
         return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
yurishkuro: this is already defined on the OpenTracing span that we extend. Why do we need to override it instead of doing putting this logic into OT?

yurishkuro:
closing for inactivity

PR #160, opened 2018-04-19
Add IP tag to tracer tags

black-adder:
Previously, the python client was using zipkin thrift to send spans to the agent. We changed this to use jaeger.thrift but lost the process IP tag in the process (<-pun). https://github.com/jaegertracing/jaeger-client-python/pull/111/files#diff-e57d8dbbca5079e5994623201c15401cL127 <- We added the IP to the zipkin endpoint object and it was converted to a process IP tag by the collector. Once we removed zipkin, we forgot to add back setting the IP tag.

### DIFF
--- a/jaeger_client/constants.py
+++ b/jaeger_client/constants.py
@@ -47,6 +47,9 @@ JAEGER_VERSION_TAG_KEY = 'jaeger.version'
 # Tracer-scoped tag that contains the hostname
 JAEGER_HOSTNAME_TAG_KEY = 'jaeger.hostname'
 
+# Tracer-scoped tag that is used to report ip of the process.
+JAEGER_IP_TAG_KEY = 'ip'
+
 # the type of sampler that always makes the same decision.
 SAMPLER_TYPE_CONST = 'const'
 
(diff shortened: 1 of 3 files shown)


black-adder:
the coverage decrease is flaky, I'm innocent

yurishkuro:
Can you book a ticket for the flaky test?

jaeger_client/constants.py
```diff
@@ -47,6 +47,9 @@
 # Tracer-scoped tag that contains the hostname
 JAEGER_HOSTNAME_TAG_KEY = 'jaeger.hostname'
```
yurishkuro: Does this need to have jaeger. prefix?

black-adder: Go, Cpp, and Java use `hostname` and this and node use `jaeger.hostname`. I can go with `hostname`

yurishkuro: let's change

yurishkuro: `jaeger.hostname` made sense with Zipkin model when it was transferred to span tags, but in the `Process` the prefix is not necessary.

PR #162, opened 2018-04-20
Preparing release 3.9.0

black-adder:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,11 +3,15 @@
 History
 -------
 
-3.9.0 (unreleased)
+3.9.0 (2018-04-20)
 ------------------
 
 - Add ability to validate configuration (#124) <Gregory Reshetniak>
 - Make Metrics consistent with Go client (#129) <Eundoo Song>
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -6,12 +6,16 @@ History
 3.9.0 (2018-04-20)
 ------------------
 
+Python 3.6 is now officially supported!
+
+- Remove support for non-ascii baggage keys; enable testing with Py 3.6  (#154) <Yuri Shkuro>
```
black-adder: Not a big fan of including names of people actually on the team, we should be celebrating our contributors, not ourselves. Maybe just me

yurishkuro: Don't have a strong opinion, but tend to agree. If we want to thanks the contributors, I prefer `thanks @kbroughton` format (as in 3.8.0), because it keeps permanent links to their Github pages. But it's more work to create the changelog since git history doesn't have screen names, only full names.

black-adder: I think people rather have their full names here, but let's just land this and deploy and change it for next time.

PR #167, opened 2018-04-29
Allow specifying hostname and ip via tags

### LINKED ISSUE #166: Ability to specify custom `jaeger.hostname` useful for running in kubernetes or containers
This code:
https://github.com/jaegertracing/jaeger-client-python/blob/34364d06e7516384a51551f80478519a51fd13a3/jaeger_client/tracer.py#L89-L90

Is not ideal when running `jaeger_client` in containerized applications. It seems most useful to allow setting a custom hostname in the config or via an environment variable. We're currently using an `init_tracer()` like this for flask applications:

```python
def initialize_tracer():
    config = Config(
        config={
            'sampler': {
                'type': 'const',
                'param': 1,
            },
            'local_agent': {
                'reporting_host': os.environ.get('NODE_NAME', '127.0.0.1'),
            },
            'logging': True,
        },
        service_name=os.environ.get('APP_DEPLOYMENT_NAME', __name__),
    )
    return config.initialize_tracer()
```

The `NODE_NAME` env var is set via the [kubernetes downward api](https://github.com/kubernetes/website/blob/master/docs/tasks/inject-data-application/dapi-envars-pod.yaml#L17-L21) to the actual kubelet running the pod (since it runs a jaeger-agent pod on every node).

Changing the code for this seems pretty trivial, but what do you suggest we name the config and env variable for this?

SEJeff:
This is much more useful in containerized environments such as kubernetes where the hostname and ip are generally useless autogenerated data. In kubernetes, the node name [can be injected](https://github.com/kubernetes/website/blob/8fd155a15f17181e65fafd051d485237ec0bd132/docs/tasks/inject-data-application/dapi-envars-pod.yaml#L18-L21) into the container as an env variable to be used.

I also cleaned up the bare except in the hostname lookup as [socket.error](https://docs.python.org/2/library/socket.html#socket.error) is the base exception for anything the module raises.

This change makes the python client match the java client from a logic perspective see:
jaegertracing/jaeger-client-java#371

Fixes #166

### DIFF
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -85,11 +85,13 @@ class Tracer(opentracing.Tracer):
         if tags:
             self.tags.update(tags)
         # noinspection PyBroadException
-        try:
-            hostname = socket.gethostname()
-            self.tags[constants.JAEGER_HOSTNAME_TAG_KEY] = hostname
-        except:
-            logger.exception('Unable to determine host name')
+
+        if self.tags.get(constants.JAEGER_HOSTNAME_TAG_KEY) is None:
+            try:
+                hostname = socket.gethostname()
+                self.tags[constants.JAEGER_HOSTNAME_TAG_KEY] = hostname
+            except:
+                logger.exception('Unable to determine host name')
 
         self.reporter.set_process(
             service_name=self.service_name,
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -29,7 +29,6 @@ from . import constants
 from .codecs import TextCodec, ZipkinCodec, ZipkinSpanFormat, BinaryCodec
 from .span import Span, SAMPLED_FLAG, DEBUG_FLAG
 from .span_context import SpanContext
-from .thrift import ipv4_to_int
 from .metrics import Metrics, LegacyMetricsFactory
 from .utils import local_ip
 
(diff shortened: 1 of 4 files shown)


jaeger_client/tracer.py
```diff
@@ -80,16 +79,19 @@ def __init__(
...
             self.tags.update(tags)
-        # noinspection PyBroadException
-        try:
-            hostname = socket.gethostname()
-            self.tags[constants.JAEGER_HOSTNAME_TAG_KEY] = hostname
-        except:
-            logger.exception('Unable to determine host name')
+
+        if self.tags.get(constants.JAEGER_IP_TAG_KEY) is None:
+            self.tags[constants.JAEGER_IP_TAG_KEY] = ipv4_to_int(local_ip())
```
yurishkuro: I suggest we remove `ipv4_to_int` portion while at it - it made sense in Zipkin format which had IP as an int field, but Jaeger converts all tags to strings, so this is no longer useful (in fact we have enricher in the backend that does the reverse conversion). Also, with string we can support ipv6.

SEJeff:
@yurishkuro I didn't see `ipv4_to_int()` used anywhere else sans tests, so it has been duly purged with :fire:

yurishkuro:
thanks!

PR #172, opened 2018-05-04
Use String strategyType instead of int

### LINKED ISSUE #171: Remote sampler always fails to refresh

The remote sampler fails to refresh and always uses the default sampler

It looks like there are two sampling endpoints on `jaeger-agent`:
`/sampling` and `/`. 
The former returns the strategy as a string (e.g. `{"strategyType":"PROBABILISTIC"}`), whereas the latter returns it as an integer (e.g. `{"strategyType":0}`)

The python client appears to be expecting the integer version, but it is querying the `/sampling` endpoint. This causes the following error when refreshing the sampling strategy:
```
  File ".../jaeger_client/sampler.py", line 438, in _update_sampler
    self._update_rate_limiting_or_probabilistic_sampler(response)
  File ".../jaeger_client/sampler.py", line 470, in _update_rate_limiting_or_probabilistic_sampler
    raise ValueError('Unsupported sampling strategy type: %s' % s_type)
ValueError: Unsupported sampling strategy type: PROBABILISTIC

```

We should use string constants for comparison here: https://github.com/jaegertracing/jaeger-client-python/blob/master/jaeger_client/sampler.py#L452

black-adder:
Resolves #171 

Unfortunately, when we swapped over to using string type for thrift instead of int for the python client, I forgot to swap over for the vanilla samplers.

I'm pretty sure our python clients are internally hooked up to emit logs on failure so I'm not sure why this wasn't reported by anyone. We should create some integration tests that handles this.

### DIFF
--- a/jaeger_client/sampler.py
+++ b/jaeger_client/sampler.py
@@ -33,9 +33,6 @@ from .constants import (
 from .metrics import Metrics, LegacyMetricsFactory
 from .utils import ErrorReporter
 from .rate_limiter import RateLimiter
-from jaeger_client.thrift_gen.sampling import (
-    SamplingManager
-)
 
 default_logger = logging.getLogger('jaeger_tracing')
 
@@ -55,6 +52,8 @@ OPERATION_SAMPLING_STR = 'operationSampling'
 MAX_TRACES_PER_SECOND_STR = 'maxTracesPerSecond'
 RATE_LIMITING_SAMPLING_STR = 'rateLimitingSampling'
 STRATEGY_TYPE_STR = 'strategyType'
+PROBABILISTIC_SAMPLING_STRATEGY = 'PROBABILISTIC'
+RATE_LIMITING_SAMPLING_STRATEGY = 'RATE_LIMITING'
 
 
 class Sampler(object):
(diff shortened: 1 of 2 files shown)


jaeger_client/sampler.py
```diff
@@ -449,10 +448,10 @@ def _update_adaptive_sampler(self, per_operation_strategies):
 
     def _update_rate_limiting_or_probabilistic_sampler(self, response):
         s_type = response.get(STRATEGY_TYPE_STR)
-        if s_type == SamplingManager.SamplingStrategyType.PROBABILISTIC:
+        if s_type == PROBABILISTIC_SAMPLING_STRATEGY:
```
yurishkuro: we could also do `|| s_type == SamplingManager.SamplingStrategyType.PROBABILISTIC`

yurishkuro:
Could you describe how you tested this?  I was under impression that the test_driver had a test for ensuring the clients read the appropriate sampling policy (but maybe it's only in the internal adaptive sampling version).

yurishkuro:
Ticket for integration tests: https://github.com/jaegertracing/jaeger/issues/804

black-adder:
Started all in one which I configured with static sampling strategies. Observed it failing and then with the change, it successfully used the new strategies.

PR #179, opened 2018-05-24
Use items instead of iteritems on Python 3

pslacerda:
The generated thrift was using `__dict__.iteritems()` even on Python 3, so the awk script was updated to handle both Python 2 and 3.

Also the generated thrift output was without the user id set.

### DIFF
--- a/thrift-gen-fix.awk
+++ b/thrift-gen-fix.awk
@@ -1,14 +1,15 @@
-BEGIN {six=0; xrange=0}
+BEGIN {six=0}
 
 /^$/ {
-    # if in the future we need six, it can be added with the following line
-    # if (six == 0) print "import six"
+    if (six == 0) {
+        print "import six";
+        print "from six.moves import xrange";
+    }
     six = 1
-    if (xrange ==0) print "from six.moves import xrange"
-    xrange = 1
 }
 
 {
-        gsub(/from ttype/, "from .ttype",$0); 
-        print
+    gsub(/from ttype/, "from .ttype", $0);
+    gsub(/self.__dict__.iteritems\(\)/, "six.iteritems(self.__dict__)", $0);
+    print
 }
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ THRIFT_GEN_DIR=jaeger_client/thrift_gen
 THRIFT_VER=0.9.3
 THRIFT_IMG=thrift:$(THRIFT_VER)
 THRIFT_PY_ARGS=new_style,tornado
-THRIFT=docker run -v "${PWD}:/data" $(THRIFT_IMG) thrift
+THRIFT=docker run -v "${PWD}:/data" -u $(shell id -u) $(THRIFT_IMG) thrift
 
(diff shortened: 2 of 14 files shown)


pslacerda:
Just for the note that because some bug on Github or Coveralls I received an email saying that the coverage was the same after this pull.

image

pslacerda:
I would keep  `__repr__` methods untested, but if it is really wanted, would suffice to do something like this:

```
def test_tag_repr(tracer):
    from jaeger_client.thrift_gen.jaeger.ttypes import Tag
    span = tracer.start_span("test-span")
    span.set_tag('python', 'is great')
    assert eval(repr(span.tags)) == span.tags
```

And so on for other generated Thrift classes.

thrift-gen-fix.awk
```diff
@@ -9,6 +9,11 @@ BEGIN {six=0; xrange=0}
 }
 
 {
-        gsub(/from ttype/, "from .ttype",$0); 
-        print
+    gsub(/from ttype/, "from .ttype", $0);
+    gsub( \
+        /self.__dict__.iteritems\(\)/, \
+        "getattr(self.__dict__, 'iteritems', self.__dict__.items)()", \
```
yurishkuro: looks a bit complicated, I would rather use the helper function from `six` module, just like we do in the main code.

pslacerda: Then maybe the following could be better.

```awk
BEGIN {six=0}

/^$/ {
    if (six == 0) {
        print "from six import iteritems";
        print "from six.moves import xrange";
    }
    six = 1
}

{
    gsub(/from ttype/, "from .ttype", $0);
    gsub(/self.__dict__.iteritems\(\)/, "iteritems(self.__dict__.items)", $0);
    print
}
```

pslacerda:
I wonder why thrift don't already use `six.iteritems()`.

yurishkuro:
I would prefer to follow the pattern we use elsewhere: `six.iteritems()`:
```
$ grep -r iteritem jaeger_client | grep -v thrift_gen
jaeger_client/codecs.py:            for key, value in six.iteritems(baggage):
jaeger_client/codecs.py:        for key, value in six.iteritems(carrier):
jaeger_client/thrift.py:        for k, v in six.iteritems(tags or {})
jaeger_client/sampler.py:        for _, sampler in six.iteritems(self.samplers):
jaeger_client/span.py:            for k, v in six.iteritems(tags):
jaeger_client/tracer.py:                    for k, v in six.iteritems(sampler_tags):
```

yurishkuro:
Please make sure your commits are signed (see CONTRIBUTING.md)

pslacerda:
Do you use squashing? If so, then when it is done I want to squash by myself to preserve my signature.

pslacerda:
Maybe could be nice to generate both to `--gen py` and `--gen py3` and put in different modules and import conditionally. I don't know what you think but can also be a nice alternative even if a bit more bloated.

yurishkuro:
thanks!

PR #182, opened 2018-05-29
Fix handling of missing ParentSpanId for B3 codec

gravelg:
There was a bug handling the trace headers when the ParentSpanId header was missing

Signed-off-by: Gabriel Gravel <ggravel@appnexus.com>


yurishkuro:
Thanks for the fix!

PR #176, opened 2018-05-20
Refactor for future span sending over http

ProvoK:
Related to https://github.com/jaegertracing/jaeger-client-python/issues/98
@yurishkuro 

> Ideally you should not need to change in the crossdock code, because if you do it means your change is not backwards compatible.

I changed it because `Reporter` and `RemoteControlledSampler` init changes. 

EDIT: reverted changes on `RemoteControlledSampler` and make `io_loop` parameter not required in `Sender` classes (`LocalAgentSender` handles io_loop=None). This will impact less actual existing codebase (and tests)

PR #184, opened 2018-05-30
Revert 'Refactor for future span sending over http'

black-adder:
Reverts jaegertracing/jaeger-client-python#176

reverting breaking change


black-adder:
@ProvoK I'm reverting it for now, it introduced a breaking change. I'll look at the PR and see how we can move forward

ProvoK:
@black-adder I might know how to solve it. But it will make the codebase even less clear, and the interface of the class (reporter) too.
As for now, imho, `Reporter` was not intended to work with anything except a local agent, that's why `channel` is passed directly, and makes this change a breaking one

PR #186, opened 2018-05-31
Preliminary refactor for supporting spans over HTTP

ProvoK:
`Reporter` now delegate span sending to new `Sender` classes .

Related to https://github.com/jaegertracing/jaeger-client-python/pull/176 but made backwards compatible.

I'm not sure this is the best approach to solve backwards compatibility though

Signed-off-by: vitto <vittorio.camisa@gmail.com>

### DIFF
--- a/jaeger_client/reporter.py
+++ b/jaeger_client/reporter.py
@@ -26,10 +26,9 @@ from .constants import DEFAULT_FLUSH_INTERVAL
 from . import thrift
 from . import ioloop_util
 from .metrics import Metrics, LegacyMetricsFactory
+from .senders import UDPSender
 from .utils import ErrorReporter
 
-from thrift.protocol import TCompactProtocol
-from jaeger_client.thrift_gen.agent import Agent
 
 default_logger = logging.getLogger('jaeger_tracing')
 
@@ -78,9 +77,11 @@ class Reporter(NullReporter):
     def __init__(self, channel, queue_capacity=100, batch_size=10,
                  flush_interval=DEFAULT_FLUSH_INTERVAL, io_loop=None,
                  error_reporter=None, metrics=None, metrics_factory=None,
-                 **kwargs):
+                 sender=None, **kwargs):
         """
         :param channel: a communication channel to jaeger-agent
+        :param sender: senders.Sender subclass implementing send method,
+            for sending batch of spans to jaeger.
         :param queue_capacity: how many spans we can hold in memory before
             starting to drop spans
         :param batch_size: how many spans we can submit at once to Collector
@@ -97,19 +98,20 @@ class Reporter(NullReporter):
         """
         from threading import Lock
 
-        self._channel = channel
+        # TODO for next major rev: remove channel param in favor of sender
+        self._sender = sender or self._create_default_sender(channel)
         self.queue_capacity = queue_capacity
         self.batch_size = batch_size
         self.metrics_factory = metrics_factory or LegacyMetricsFactory(metrics or Metrics())
         self.metrics = ReporterMetrics(self.metrics_factory)
         self.error_reporter = error_reporter or ErrorReporter(Metrics())
         self.logger = kwargs.get('logger', default_logger)
-        self.agent = Agent.Client(self._channel, self)
 
         if queue_capacity < batch_size:
             raise ValueError('Queue capacity cannot be less than batch size')
 
-        self.io_loop = io_loop or channel.io_loop
+        self.io_loop = io_loop or self.fetch_io_loop(channel, self._sender)
+
         if self.io_loop is None:
             self.logger.error('Jaeger Reporter has no IOLoop')
         else:
@@ -122,6 +124,22 @@ class Reporter(NullReporter):
 
         self._process_lock = Lock()
         self._process = None
+
+    @staticmethod
+    def fetch_io_loop(channel, sender):
+        if channel:
+            return channel.io_loop
+        elif sender:
+            return sender.io_loop
+        return None
+
+    def _create_default_sender(self, channel):
+        sender = UDPSender(
+            port=channel.reporting_port,
+            host=channel.host,
+            io_loop=channel.io_loop
+        )
+        return sender
 
     def set_process(self, service_name, tags, max_length):
         with self._process_lock:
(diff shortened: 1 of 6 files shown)


jaeger_client/reporter.py
```diff
@@ -75,12 +74,14 @@ def report_span(self, span):
 
 class Reporter(NullReporter):
     """Receives completed spans from Tracer and submits them out of process."""
-    def __init__(self, channel, queue_capacity=100, batch_size=10,
+    def __init__(self, channel, sender=None, queue_capacity=100, batch_size=10,
```
black-adder: shouldn't the new `sender` parameter be the last parameter since users could be calling this function without explicitly naming(?) the parameters?

yurishkuro: and add a TODO to fix that in the next major rev.

ProvoK: @black-adder you're right! Should have been fixed with also the todo @yurishkuro asked for.
Is it maybe a good thing open the issue for that and assign it to the v4 milestone (if it exists)?

 Let me know is anything else is necessary :)

jaeger_client/reporter.py
```diff
@@ -78,9 +77,11 @@ class Reporter(NullReporter):
     def __init__(self, channel, queue_capacity=100, batch_size=10,
                  flush_interval=DEFAULT_FLUSH_INTERVAL, io_loop=None,
                  error_reporter=None, metrics=None, metrics_factory=None,
-                 **kwargs):
+                 sender=None, **kwargs):
```
black-adder: would making sender a `kwargs` be the safest bet? Won't this still run into the issue where sender might be incorrectly interpreted as kwargs?

yurishkuro: I think `sender=None, **kwargs` is good

ProvoK: @black-adder I think that putting `sender` in `kwargs` will hid it more than necessary, without any benefit.
Honestly, if a developer just writes down 8/9 positional arguments, all without naming (aka kwarg), he actually deserves some kind of error :laughing:

yurishkuro: I agree

jaeger_client/reporter.py
```diff
@@ -97,19 +98,19 @@ def __init__(self, channel, queue_capacity=100, batch_size=10,
...
         self.metrics = ReporterMetrics(self.metrics_factory)
         self.error_reporter = error_reporter or ErrorReporter(Metrics())
         self.logger = kwargs.get('logger', default_logger)
-        self.agent = Agent.Client(self._channel, self)
 
         if queue_capacity < batch_size:
             raise ValueError('Queue capacity cannot be less than batch size')
 
-        self.io_loop = io_loop or channel.io_loop
+        self.io_loop = io_loop or channel.io_loop or self.sender.io_loop
```
yurishkuro: if `channel is None` this will blow up

jaeger_client/reporter.py
```diff
@@ -97,19 +98,19 @@ def __init__(self, channel, queue_capacity=100, batch_size=10,
         """
         from threading import Lock
 
-        self._channel = channel
+        # TODO for next major rev: remove channel param in favor of sender
+        self.sender = sender or self._create_default_sender(channel)
```
yurishkuro: make it private, `_sender`

jaeger_client/reporter.py
```diff
@@ -213,7 +213,7 @@ def _send(self, batch):
         Send batch of spans out via thrift transport. Any exceptions thrown
         will be caught above in the exception handler of _submit().
         """
-        return self.agent.emitBatch(batch)
+        return self.sender.send(batch)
```
yurishkuro: This is a breaking change it the user still passes the channel. There needs to be an adapter to wrap the old channel and make it look like a sender with `send()` method.

ProvoK: Actually, if the user passes the channel, the channel itself is used to generate a default UDPSender (`_create_default_sender(channel)`). 
I dont understand were is here the breaking change

jaeger_client/reporter.py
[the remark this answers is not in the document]
ProvoK: I'm gonna make a test case with this fails scenario (sender not None, but io_loop and channel None) and then fix the code. Sorry I missed that

jaeger_client/reporter.py
yurishkuro: +1

jaeger_client/reporter.py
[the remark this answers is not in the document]
ProvoK: I did it, and I fixed the linting problem that broke the pipeline.
About your open question, I think that you are right, if there are those differences it's better that the sender should to that job.
In that case, I really should look at the code and do another "refactor".

ProvoK:
Hello there,
@rmfitzpatrick was very kind and continued the work and PR to me the changes for adopting what I think @yurishkuro asked for.

jaeger_client/reporter.py
```diff
@@ -78,9 +75,11 @@ class Reporter(NullReporter):
     def __init__(self, channel, queue_capacity=100, batch_size=10,
```
yurishkuro: should this be `channel=None`?

yurishkuro: nvm it's still required

rmfitzpatrick:
@yurishkuro, has there potentially been any update on this review?

black-adder:
@rmfitzpatrick sorry about the delay, I'll take a look at this today

jaeger_client/senders.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @property
+    def span_count(self):
+        return len(self.spans)
+
+    @tornado.gen.coroutine
+    def flush(self):
+        """Examine span and process state before yielding to _flush() for batching and transport."""
+        if self.spans:
+            with self._process_lock:
```
black-adder: for another PR, but the locking here seems a little weird looking at it now. I wonder why we only lock on the process but skip locking on the spans which could be flushed and emptied in another thread.

rmfitzpatrick:
At @ProvoK's request, I'm attempting to carry the baton to https://github.com/jaegertracing/jaeger-client-python/pull/208 to continue this effort.

shuaichang:
> #208

Just curious, is this PR still an active effort?

PR #188, opened 2018-06-18
Implement throttler

isaachier:


isaachier:
I'll work on signing the commits at some later point.

crossdock/server/endtoend.py
```diff
@@ -101,10 +106,19 @@ def tracers(self, tracers):
...
+        try:
+            host, port_str = os.getenv('AGENT_HOST_PORT').split(':')
+            port = int(port_str)
+        except:
+            host = 'jaeger-agent'
+            port = DEFAULT_REPORTING_PORT
+            pass
+
         return LocalAgentSender(
-            host=os.getenv('AGENT_HOST', 'jaeger-agent'),
```
black-adder: so this was never used?

jaeger_client/local_agent_net.py
```diff
@@ -51,15 +62,19 @@ class LocalAgentSender(TBufferedTransport):
...
-    def __init__(self, host, sampling_port, reporting_port, io_loop=None):
+    def __init__(self, host, sampling_port, reporting_port, io_loop=None, throttling_port=None):
         # IOLoop
         self._thread_loop = None
         self.io_loop = io_loop or self._create_new_thread_loop()
 
-        # http sampling
-        self.local_agent_http = LocalAgentHTTP(host, sampling_port)
+        # HTTP sampling
+        self.sampling_http = LocalAgentHTTP(host, sampling_port)
```
black-adder: this is technically still a public field, let's keep the old name for now

jaeger_client/tracer.py
```diff
@@ -230,6 +235,11 @@ def report_span(self, span):
     def random_id(self):
         return self.random.getrandbits(constants.MAX_ID_BITS)
 
+    def is_allowed(self, *args, **kwargs):
```
black-adder: this function name makes no sense in this context

jaeger_client/span.py
```diff
@@ -78,12 +78,9 @@ def set_tag(self, key, value):
         :param value:
         """
         with self.update_lock:
-            if key == ext_tags.SAMPLING_PRIORITY:
-                if value > 0:
-                    self.context.flags |= SAMPLED_FLAG | DEBUG_FLAG
-                else:
-                    self.context.flags &= ~SAMPLED_FLAG
-            elif self.is_sampled():
+            if key == ext_tags.SAMPLING_PRIORITY and self._set_sampling_priority(value):
```
black-adder: if the priority flag setting was successful, shouldn't we add the tag? Here it looks like we skip on success.

jaeger_client/span.py
```diff
@@ -92,6 +89,19 @@ def set_tag(self, key, value):
                 self.tags.append(tag)
         return self
 
+    def _set_sampling_priority(self, value):
```
black-adder: joe made a good point in the node PR; if the span is already debug sampled, setting the tag again should NOT call is_allowed as that can decrement the credits and leave the client and the agent out of sync

jaeger_client/local_agent_net.py
isaachier: Sure. With Python I find it difficult to distinguish public and prive.

crossdock/server/endtoend.py
isaachier: I believe it is the wrong `env` variable. It should be `JAEGER_AGENT_HOST`. I will double check this.

jaeger_client/tracer.py
isaachier: True. I will use the name found in the Go client (`isDebugAllowed`), which is more clear. Will fix this.

jaeger_client/span.py
isaachier: Wow good catch. I must have messed this up. I'm surprised it still seems to work in the crossdock test. Then again, the propagation there might act differently.

jaeger_client/span.py
isaachier: Definitely. We should probably update the Go client too at some point with this fix.

crossdock/server/endtoend.py
```diff
@@ -101,10 +106,19 @@ def tracers(self, tracers):
 
     @property
     def local_agent_sender(self):
+        try:
+            host, port_str = os.getenv('AGENT_HOST_PORT').split(':')
```
yurishkuro: this is rather dirty. If there's a legitimate error, like port not a number, it should be logged. But the way this try/catch is structured it can blow even when the env var is empty.

crossdock/server/endtoend.py
isaachier: BTW seems to be only for crossdock, not a general client use case.

isaachier: Definitely never used in either Dockerfile not docker-compose.yml.

crossdock/server/endtoend.py
isaachier: Ya I don't really like it but this config module does stuff like this all over the place. I'll see what I can do.

jaeger_client/tracer.py
```diff
@@ -92,6 +93,10 @@ def __init__(
             except socket.error:
                 logger.exception('Unable to determine host name')
 
+        self.throttler = throttler
+        if self.throttler:
+            self.tags[constants.CLIENT_UUID_TAG_KEY] = self.throttler.uuid
```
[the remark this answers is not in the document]
isaachier: Makes sense.

isaachier:
Hey @black-adder @yurishkuro how's it looking now?

jaeger_client/throttler.py
```diff
@@ -0,0 +1,199 @@
...
+
+            self.credits[operation] = value - MINIMUM_CREDITS
+            return True
+
+    def _set_client_id(self, client_id):
+        """
+        Method for tracer to set client ID of throttler.
+        """
+        with self.lock:
+            if self.client_id is None:
```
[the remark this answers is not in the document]
isaachier: Are you asking or suggesting? I did lock here because to make the interface completely thread-safe. Seeing as the caller is the tracer constructor, I don't really foresee a race condition. What do you think is more correct?

jaeger_client/throttler.py
```diff
@@ -0,0 +1,199 @@
...
+            self.metrics.throttler_update_failure(1)
+            self.error_reporter.error(
+                'Failed to get throttling credits from jaeger-agent: %s',
+                exception)
+            return
+
+        response = future.result()
+        # In Python 3.5 response.body is of type bytes and json.loads() does only support str
+        # See: https://github.com/jaegertracing/jaeger-client-python/issues/180
+        if hasattr(response.body, 'decode') and callable(response.body.decode):
```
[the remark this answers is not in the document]
isaachier: This is almost entirely copied from the sampler.

tests/test_throttler.py
```diff
@@ -0,0 +1,95 @@
...
+import mock
+import time
+
+from jaeger_client.throttler import RemoteThrottler
+
+
+def test_throttler_simple():
+    channel = mock.MagicMock()
+    throttler = RemoteThrottler(channel, 'test-service')
+    allowed = throttler.is_allowed('test-operation')
```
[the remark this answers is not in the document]
isaachier: Sure. I guess this is a bad habit from C/C++. Common advice there is not to evaluate expressions that have side-effects inside an assertion because it will be omitted from optimized builds. I'll try to "un-learn" this advice for Python.

black-adder:
one last thing, we need to take into account of debug headers: https://github.com/isaachier/jaeger-client-python/blob/ceae3df8f51b4fa5e2be0fe8848d9a4eaf509cf0/jaeger_client/tracer.py#L161

they should also be throttled

isaachier:
Anything else?

isaachier:
Thanks!

PR #192, opened 2018-06-26
Improve handling of config values and default arguments

isaachier:
Fix for:
```
File "/usr/local/lib/python2.7/dist-packages/jaeger_client/sampler.py", line 312, in __str__
    self.max_operations)
TypeError: %d format: a number is required, not NoneType
Logged from file sampler.py, line 428
```

Signed-off-by: Isaac Hier <ihier@uber.com>

### DIFF
--- a/jaeger_client/sampler.py
+++ b/jaeger_client/sampler.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2016 Uber Technologies, Inc.
+# Copyright (c) 2016-2018 Uber Technologies, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -264,7 +264,7 @@ class AdaptiveSampler(Sampler):
         self.max_operations = max_operations
 
     def is_sampled(self, trace_id, operation=''):
-        sampler = self.samplers.get(operation, None)
+        sampler = self.samplers.get(operation)
         if sampler is None:
             if len(self.samplers) >= self.max_operations:
                 return self.default_sampler.is_sampled(trace_id, operation)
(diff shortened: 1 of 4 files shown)


jaeger_client/sampler.py
```diff
@@ -264,7 +264,7 @@ def __init__(self, strategies, max_operations):
         self.max_operations = max_operations
 
     def is_sampled(self, trace_id, operation=''):
-        sampler = self.samplers.get(operation, None)
+        sampler = self.samplers.get(operation)
         if sampler is None:
```
yurishkuro: a) can you add a test reproducing the exact exception the users got? I added it to the description.
b) should the above be `if sampler:` instead ? Otherwise I don't see what's really changing here.

jaeger_client/sampler.py
```diff
@@ -264,8 +264,8 @@ def __init__(self, strategies, max_operations):
         self.max_operations = max_operations
 
     def is_sampled(self, trace_id, operation=''):
-        sampler = self.samplers.get(operation, None)
```
yurishkuro: I thought the reason we pass the default value is because the dictionary might otherwise blow up if the key is not present. Or is it only if we use brackets `sampler[operation]`?

In any case, I don't think that the get() statement itself needs to change, according to your explanation it was more that the `if X is not None` checks that were tripping the error.

isaachier: Actually Python defaults to `None` if no second argument is passed so this is just unnecessary (see https://docs.python.org/2/library/stdtypes.html#dict.get). I fixed all places in these files where I found that. If you want me to revert the unrelated changes I can.

yurishkuro: nah, it's fine

isaachier: :)

isaachier:
OK so my first PR (#188) has the fixes for the py.test issues that are currently breaking Travis. Will check them out and add them here.

isaachier:
coveralls is annoying. Need to rerun the build somehow.

isaachier:
Even rerunning this build, coveralls is not working. Here is the link to the coverage report: https://coveralls.io/builds/17718966.

yurishkuro:
do you want to create another PR and switch to codecov? then rebase this.

Cf: https://github.com/jaegertracing/jaeger-client-node/pull/269

PR #194, opened 2018-06-27
Move to codecov

isaachier:


isaachier:
@yurishkuro are you okay with this change?

PR #196, opened 2018-07-02
Add ability to update rate limiter

black-adder:

- The current way the rate limiting sampler is updated involves the sampler being reconstructed; this has the unfortunate side effect of the balance being reinitialized to full which will generate a new sample. If you are updating thousands of instances with a new rate limiting rate, this will generate thousands of new traces. This PR adds the ability to update the rate limiter so that the proportional balance is retained on update.

cf. https://github.com/jaegertracing/jaeger-client-go/pull/320

### DIFF
--- a/jaeger_client/sampler.py
+++ b/jaeger_client/sampler.py
@@ -143,19 +143,26 @@ class RateLimitingSampler(Sampler):
     """
 
     def __init__(self, max_traces_per_second=10):
-        super(RateLimitingSampler, self).__init__(
-            tags={
-                SAMPLER_TYPE_TAG_KEY: SAMPLER_TYPE_RATE_LIMITING,
-                SAMPLER_PARAM_TAG_KEY: max_traces_per_second,
-            }
-        )
+        super(RateLimitingSampler, self).__init__()
+        self.rate_limiter = None
+        self._init(max_traces_per_second)
+
+    def _init(self, max_traces_per_second):
         assert max_traces_per_second >= 0, \
             'max_traces_per_second must not be negative'
+        self._tags = {
+            SAMPLER_TYPE_TAG_KEY: SAMPLER_TYPE_RATE_LIMITING,
+            SAMPLER_PARAM_TAG_KEY: max_traces_per_second,
+        }
         self.traces_per_second = max_traces_per_second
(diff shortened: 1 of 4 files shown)


jaeger_client/sampler.py
```diff
@@ -143,19 +143,26 @@ class RateLimitingSampler(Sampler):
...
+        self._tags = {
+            SAMPLER_TYPE_TAG_KEY: SAMPLER_TYPE_RATE_LIMITING,
+            SAMPLER_PARAM_TAG_KEY: max_traces_per_second,
+        }
         self.traces_per_second = max_traces_per_second
-        self.rate_limiter = RateLimiter(
-            credits_per_second=self.traces_per_second,
-            max_balance=self.traces_per_second if self.traces_per_second > 1.0 else 1.0
-        )
+        max_balance = self.traces_per_second if self.traces_per_second > 1.0 else 1.0
```
isaachier: I find the built-in function `max` to be more readable here.

jaeger_client/sampler.py
```diff
@@ -143,19 +143,26 @@ class RateLimitingSampler(Sampler):
...
+        if not self.rate_limiter:
+            self.rate_limiter = RateLimiter(
+                credits_per_second=self.traces_per_second,
+                max_balance=max_balance
```
isaachier: You can use the `max` function inline here to be consistent with other clients.

jaeger_client/sampler.py
```diff
@@ -458,16 +471,20 @@ def _update_adaptive_sampler(self, per_operation_strategies):
...
             sampling_rate = get_sampling_probability(response)
             new_sampler = ProbabilisticSampler(rate=sampling_rate)
         elif s_type == RATE_LIMITING_SAMPLING_STRATEGY:
             mtps = get_rate_limit(response)
-            if 0 <= mtps < 500:
-                new_sampler = RateLimitingSampler(max_traces_per_second=mtps)
-            else:
+            if mtps < 0 or mtps >= 500:
                 raise ValueError(
                     'Rate limiting parameter not in [0, 500] range: %s' % mtps)
```
isaachier: Nit: if excluding `500`, you should use `[0, 500)` notation.

PR #200, opened 2018-07-13
Fix incorrect default agent reporting port

rmfitzpatrick:
Corrects the returned default reporting port that was likely an
accident.

edit: Originally I had updated the evaluation order, but the config should take priority.

Signed-off-by: Ryan Fitzpatrick <rmfitzpatrick@signalfx.com>

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -144,3 +144,7 @@ class ConfigTests(unittest.TestCase):
         tracer = c.initialize_tracer()
 
         assert opentracing.tracer == tracer
+
+    def test_default_local_agent_reporting_port(self):
+        c = Config({}, service_name='x')
+        assert c.local_agent_reporting_port == 6831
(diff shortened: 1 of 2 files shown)


tests/test_config.py
```diff
@@ -144,3 +144,7 @@ def test_initialize_tracer(self):
         tracer = c.initialize_tracer()
 
         assert opentracing.tracer == tracer
+
+    def test_default_local_agent_reporting_port(self):
+        c = Config({}, service_name='x')
+        assert c.local_agent_reporting_port == DEFAULT_REPORTING_PORT
```
black-adder: rather than testing against DEFAULT_REPORTING_PORT, can you explicitly test against the value? This way next time someone inadvertently changes DEFAULT_REPORTING_PORT to something else, this test will break.

rmfitzpatrick: Updated.

yurishkuro:
Looks like it was recently introduced https://github.com/jaegertracing/jaeger-client-python/pull/188/files#diff-02ca461860c9f795c0f4049be6926409R254

yurishkuro:
good to have the test now!

PR #202, opened 2018-07-26
This is because thrift for python doesn't have 'package_prefix'

zgfh:


yurishkuro:
I don't understand what problem this is solving.

yurishkuro:
closing for inactivity

Issue #120, opened 2018-01-30
Illegal header key with b3 headers and GRPC


stevededalus:
I'm running into some errors trying to use jaeger on grpc requests with a zipkin compatible backend and was hoping for some pointers. I believe I just need to set the `b3` propagation flag and that should set the correct headers, but when I attempt to I get the following errors:

```
E0130 00:48:37.591239271     112 call.cc:1001]               validate_metadata: {"created":"@1517273317.591206974","description":"Illegal header key","file":"src/core/lib/surface/validate_metadata.cc","file_line":42,"offset":0,"raw_bytes":"58 2d 42 33 2d 54 72 61 63 65 49 64 'X-B3-TraceId'"}
INFO:jaeger_tracing:Reporting span 4b16842cb7f8fff0:4b16842cb7f8fff0:0:1 poiservice-client./grpc.health.v1.Health/Check

  File "poiservice/scripts/poi_example_client.py", line 68, in check_health
    response = stub.Check(request)
  File "/usr/local/lib/python3.6/site-packages/grpc_opentracing/grpcext/_interceptor.py", line 35, in __call__
    invoker)
  File "/usr/local/lib/python3.6/site-packages/grpc_opentracing/_client.py", line 138, in intercept_unary
    result = invoker(request, metadata)
  File "/usr/local/lib/python3.6/site-packages/grpc_opentracing/grpcext/_interceptor.py", line 31, in invoker
    return self._base_callable(request, timeout, metadata, credentials)
  File "/usr/local/lib/python3.6/site-packages/grpc/_channel.py", line 483, in __call__
    credentials)
  File "/usr/local/lib/python3.6/site-packages/grpc/_channel.py", line 476, in _blocking
    _check_call_error(call_error, metadata)
  File "/usr/local/lib/python3.6/site-packages/grpc/_channel.py", line 86, in _check_call_error
    raise ValueError('metadata was invalid: %s' % metadata)
TypeError: not all arguments converted during string formatting
```

The raw_byte string looks suspicious and was wondering if this could be related to a python2 -> python3 change with the handling of strings in python3 as unicode and not raw byte arrays.

Client code snippet:

```
import logging
import sys
import grpc
import time

from jaeger_client import Config

from grpc_opentracing import open_tracing_client_interceptor
from grpc_opentracing.grpcext import intercept_channel

from grpc_health.v1 import health_pb2, health_pb2_grpc

def check_health(stub):
[...]

PR #203, opened 2018-07-26
Correct sampled and debug B3 headers

rmfitzpatrick:
Distinguishes between X-B3-Sampled and X-B3-Flags B3 headers on
context injection and adds case insensitivity for carrier key
checks.

Resolves: https://github.com/jaegertracing/jaeger-client-python/issues/120

Signed-off-by: Ryan Fitzpatrick <rmfitzpatrick@signalfx.com>

### DIFF
--- a/jaeger_client/codecs.py
+++ b/jaeger_client/codecs.py
@@ -257,12 +257,16 @@ class B3Codec(Codec):
     https://github.com/openzipkin/b3-propagation
 
     """
-    def __init__(self):
-        self.trace_header = 'X-B3-TraceId'
-        self.span_header = 'X-B3-SpanId'
-        self.parent_span_header = 'X-B3-ParentSpanId'
-        self.sampled_header = 'X-B3-Sampled'
-        self.flags_header = 'X-B3-Flags'
+    trace_header = 'X-B3-TraceId'
+    _trace_header_lc = trace_header.lower()
+    span_header = 'X-B3-SpanId'
+    _span_header_lc = span_header.lower()
+    parent_span_header = 'X-B3-ParentSpanId'
+    _parent_span_header_lc = parent_span_header.lower()
+    sampled_header = 'X-B3-Sampled'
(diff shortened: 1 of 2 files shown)


jaeger_client/codecs.py
```diff
@@ -271,21 +271,25 @@ def inject(self, span_context, carrier):
...
+        elif span_context.flags & SAMPLED_FLAG == SAMPLED_FLAG:
+            carrier[self.sampled_header] = '1'
 
     def extract(self, carrier):
         if not isinstance(carrier, dict):
             raise InvalidCarrierException('carrier not a dictionary')
-        trace_id = header_to_hex(carrier.get(self.trace_header.lower()))
-        span_id = header_to_hex(carrier.get(self.span_header.lower()))
-        parent_id = carrier.get(self.parent_span_header.lower(), None)
+        ic_map = dict([(k.lower(), k) for k in carrier])
```
yurishkuro: what does `ic` stand for?

jaeger_client/codecs.py
```diff
@@ -271,21 +271,25 @@ def inject(self, span_context, carrier):
...
+        trace_id = header_to_hex(carrier.get(ic_map.get(self.trace_header.lower())))
+        span_id = header_to_hex(carrier.get(ic_map.get(self.span_header.lower())))
+        parent_id = carrier.get(ic_map.get(self.parent_span_header.lower()))
```
yurishkuro: Calling `.lower()` on a constant on every request seems highly inefficient.  Why not cache these on construction?

jaeger_client/codecs.py
rmfitzpatrick: "Ignore case".  Updated to `lowercase_keys` for clarity.

jaeger_client/codecs.py
rmfitzpatrick: These changes don't introduce this practice, but good idea.  Updated to use class attributes for single initialization.

Issue #205, opened 2018-08-06
5.0.0 Design


black-adder:
## Requirement - what kind of business use case are you trying to solve?

The python client was not architected in line with the other clients. For example, adding a new HTTP sender will require an extensive surgery and a lot of technical debt to maintain backwards compatibility; in the other clients, adding a new HTTP sender was a breeze.

There are other tasks we can accomplish such as:

- Reduce the public API to prevent semver issues #185 
- Remove the `enabled` config variable #145
- Remove tornado dependency #235

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Here's a list of changes, I'll try to make this as exhaustive as possible but I may be missing some changes:

- Create a new `Sender` interface ala the [golang client](https://github.com/jaegertracing/jaeger-client-go/blob/master/transport.go) which supports just `append` and `flush` as begun [here](https://github.com/jaegertracing/jaeger-client-python/blob/fbefb070c6dce91e757fecebbe8add760fe5d620/jaeger_client/senders.py). `LocalAgentSender` should be removed entirely and assimilated into `UDPSender` (or made a private class). `Reporter` should accept a `Sender` instead of `LocalAgentSender`.
- `LocalAgentHTTP(host, sampling_port, throttling_port)` -> `RemoteConfigManager(host, config_port)`. Since sampling_port and throttling_port is always the same, we can unify them. Update `RemoteThrottler` and `RemoteControlledSampler` to accept `RemoteConfigManager` instead of `LocalAgentSender`.
-  `config.py` changes
   - Update `throttler`
      - Remove `port`
   - Remove `sampling_refresh_interval`
   - Update `sampler`
      - Add `refresh_interval`
   - Remove `enabled` #145
   - Add `remote_config`
      - Add `host`
      - Add `port`
   - Remove `reporter_batch_size`
   - Remove `reporter_flush_interval`
   - Remove `logging`
   - Rename `local_agent` -> `reporter`
      - Remove `enabled`
      - Rename `reporting_host` -> `host`
      - Remove `reporting_port` -> `port`
      - Add `batch_size`
      - Add `flush_interval`
      - Add `log_spans`
 - Remove all deprecated functions such as `span.error` and `span.info`
 - Reduce the scope of the API and make private certain fields. Fields TBD.
 - Remove tornado dependency.

I'll keep adding to this list, probably missing something.

## Any open questions to address

[...]

Issue #31, opened 2017-02-15
Make dependency on tornado/threadloop optional


philipgian:
I would like to use jaeger-client as an opentracing tracer with zipkin.
I've already completed a proof-of-concept integration, and I fully intend to release the Zipkin codec and Zipkin reporter as soon as they are ready.

My goal is to use the client on a Django/Gunicorn/Greenlet environment, so I would like to know if there is any possibility to make the tornado/threadloop dependencies optional.

Ideally, I think that since the client's architecture is so modular, the `Reporter` could be renamed to `TornadoReporter` and properly presented on the Python package (along with the local_agent) only if the tornado/threadloop modules are on the Python PATH.

yurishkuro:
I don't have objections to this approach. However, what happens to `requirements.txt` in this case? Would we need to include all different frameworks in there? Or do we put them only in test requirements and tell the users that they need to depend on the correct module?

philipgian:
The second solution was what I had in mind. We could require all the dependencies only for the tests and "recommend" to the users the right dependencies, issuing appropriate warning messages or raising an exception if one tries to use the tornado reporter without having tornado in the python path.

Another more bureaucratic approach is to create another python package just for the framework specific code.

yurishkuro:
Update: there have been other issues and discussions related to this, I think the ideal approach is to remove Tornado dependency completely and implement async span flushing in a dedicated thread. The thread should also be started lazily (e.g. only when the first span is reported), which should address two issues:
  * when running under WSGI-style frameworks, the thread will naturally be started post-fork
  * if the thread is always started asynchronously (unlike the threadloop which currently blocks), then it won't be causing the deadlocks

zachgoldstein:
I'm playing with making this work nicely with https://github.com/opentracing-contrib/python-flask. Would love to see a separate repo for framework specific niceties, like jaeger-flask, jaeger-django etc.

Side note; thanks for all the hard work!

rmfitzpatrick:
@yurishkuro, do you think a threading.Timer would be an acceptable replacement for the PeriodicCallbacks in RemoteControlledSpampler and RemoteThrottler?  I'm not sure how straightforward it would be to ensure their invocations to be post-fork.

afaict these are the only other places that require nontrivial removal of tornado, assuming thrift-generated code is 1:1 when not specifying the tornado option.

yurishkuro:
I don't know off hand what threading.timer does. Does it require a separate thread?

[...]

PR #206, opened 2018-08-08
Use opentracing 2.0

### LINKED ISSUE #199: Support OpenTracing-2.0.0

Update requirements for the openstack/requirements project

https://github.com/jaegertracing/jaeger-client-python/blob/master/setup.py#L42
Specifically the cap on opentracing.

Update jaeger-client to work with opentracing-2 as well.

mihau:
This pull request bumps `opntracing` to 2.0 and implements changes needed for api compatibility (mainly regarding scope managers). I assumed compatibility with 2.x only, but maybe it would be possible to run it with 1.x as well, not sure if that is a good idea though. 
I used `ThreadLocalScopeManager` from `opentracing` as a default scope manager.

The tests may fail because of an `opentracing` issue I fixed in https://github.com/opentracing/opentracing-python/pull/97 

Resolves #199

### DIFF
--- a/jaeger_client/config.py
+++ b/jaeger_client/config.py
@@ -84,17 +84,20 @@ class Config(object):
     _initialized_lock = threading.Lock()
 
     def __init__(self, config, metrics=None, service_name=None, metrics_factory=None,
-                 validate=False):
+                 validate=False, scope_manager=None):
         """
         :param metrics: an instance of Metrics class, or None. This parameter
             has been deprecated, please use metrics_factory instead.
         :param service_name: default service name.
             Can be overwritten by config['service_name'].
         :param metrics_factory: an instance of MetricsFactory class, or None.
+        :param scope_manager: a class implementing a scope manager, or None for
+            default (ThreadLocalScopeManager).
         """
         if validate:
             self._validate_config(config)
         self.config = config
+        self.scope_manager = scope_manager
         if get_boolean(self.config.get('metrics', True), True):
             self._metrics_factory = metrics_factory or LegacyMetricsFactory(metrics or Metrics())
         else:
@@ -413,6 +416,7 @@ class Config(object):
             max_tag_value_length=self.max_tag_value_length,
             extra_codecs=self.propagation,
             throttler=throttler,
+            scope_manager=self.scope_manager,
         )
 
     def _initialize_global_tracer(self, tracer):
--- a/jaeger_client/config.py
+++ b/jaeger_client/config.py
@@ -91,7 +91,7 @@ class Config(object):
         :param service_name: default service name.
             Can be overwritten by config['service_name'].
         :param metrics_factory: an instance of MetricsFactory class, or None.
-        :param scope_manager: a class implementing a scope manager, or None for
+        :param scope_manager: an instance of a scope manager, or None for
             default (ThreadLocalScopeManager).
         """
         if validate:
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -142,9 +142,8 @@ class Tracer(opentracing.Tracer):
         """
         parent = child_of
 
-        active_span = self.active_span
-        if parent is None and not ignore_active_span:
-            parent = active_span
+        if self.active_span is not None and not ignore_active_span:
+            parent = self.active_span
 
         if references:
             if isinstance(references, list):
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -123,7 +123,8 @@ class Tracer(opentracing.Tracer):
                    references=None,
                    tags=None,
                    start_time=None,
-                   ignore_active_span=False):
+                   ignore_active_span=False,
+    ):
         """
         Start and return a new Span representing a unit of work.
 
@@ -211,7 +212,8 @@ class Tracer(opentracing.Tracer):
                           tags=None,
                           start_time=None,
                           ignore_active_span=False,
-                          finish_on_close=True):
+                          finish_on_close=True,
+    ):
         """
         Returns a newly started and activated :class:`Scope`
 
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -112,8 +112,7 @@ class Tracer(opentracing.Tracer):
         )
 
         super(Tracer, self).__init__(
-            scope_manager=scope_manager() if scope_manager
-            else ThreadLocalScopeManager()
+            scope_manager=scope_manager or ThreadLocalScopeManager()
         )
 
     def start_span(self,
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -25,6 +25,7 @@ import six
 import opentracing
 from opentracing import Format, UnsupportedFormatException
 from opentracing.ext import tags as ext_tags
+from opentracing.scope_managers import ThreadLocalScopeManager
 
 from . import constants
 from .codecs import TextCodec, ZipkinCodec, ZipkinSpanFormat, BinaryCodec
@@ -51,6 +52,7 @@ class Tracer(opentracing.Tracer):
         tags=None,
         max_tag_value_length=constants.MAX_TAG_VALUE_LENGTH,
         throttler=None,
+        scope_manager=ThreadLocalScopeManager,
     ):
         self.service_name = service_name
         self.reporter = reporter
@@ -87,6 +89,7 @@ class Tracer(opentracing.Tracer):
         if tags:
             self.tags.update(tags)
 
+
         if self.tags.get(constants.JAEGER_IP_TAG_KEY) is None:
             self.tags[constants.JAEGER_IP_TAG_KEY] = local_ip()
 
@@ -109,12 +112,15 @@ class Tracer(opentracing.Tracer):
             max_length=self.max_tag_value_length,
         )
 
+        super(Tracer, self).__init__(scope_manager=scope_manager())
+
     def start_span(self,
                    operation_name=None,
                    child_of=None,
                    references=None,
                    tags=None,
-                   start_time=None):
+                   start_time=None,
+                   ignore_active_span=False):
         """
         Start and return a new Span representing a unit of work.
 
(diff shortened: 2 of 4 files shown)


jaeger_client/tracer.py
```diff
@@ -126,10 +132,17 @@ def start_span(self,
...
+        :param ignore_active_span: an explicit flag that ignores the current
+            active :class:`Scope` and creates a root :class:`Span`
 
         :return: Returns an already-started Span instance.
         """
         parent = child_of
+
+        active_span = self.active_span
+        if parent is None and not ignore_active_span:
+            parent = active_span
```
black-adder: nit: `parent = self.active_span` and remove L142

setup.py
```diff
@@ -39,7 +39,7 @@
         'threadloop>=1,<2',
         'thrift',
         'tornado>=4.3,<5',
-        'opentracing>=1.2.2,<2',
+        'opentracing>=2.0,<2.1',
```
black-adder: why not `<3`?

jaeger_client/tracer.py
```diff
@@ -184,6 +197,46 @@ def start_span(self,
 
         return span
 
+    def start_active_span(self,
```
black-adder: This is my bad for not keeping up to tabs for `scopes`, but is it expected for the user to be aware of the concept of `scopes`, and `active spans`? I guess what I'm getting at is would it be better have the existing `start_span` function take care of doing `self.scope_manager.activate(span, finish_on_close)` so that there isn't a costly migration for users?

setup.py
prometheanfire: If it can be tested on 1.2.2 the lower bound could be kept.  This aids people migrating to the new version as well.

jaeger_client/tracer.py
mihau: Well this is a part of the opentracing api, and there are tests requiring that particular method to be in place (https://github.com/opentracing/opentracing-python/blob/9ef3e19f101774a53f316ff508ecaeaa4af741bc/opentracing/harness/api_check.py#L70), so it seems to me that the intent is for user to be aware of `scopes`. I don't think that the migration would be costly since one can either continue to pass parent spans to `start_span` and it would continue to work as it used to, or move to using the `start_active_span` and get rid of passing the parent.

setup.py
mihau: Out of cautiousness, but yeah - minors shouldn't break it so I'm fine with `<3`.

mihau: I could try testing with `1.2.2` and it would probably be doable to pass for both versions, but I think it would be confusing since the jaeger-client code would suggest that scope managers work, whereas in reality that would depend on the opentracing version you're using. I'd rather go with `>=2.0,<3.0`.

prometheanfire: In that case keep in mind that you will not be co-installable with your previous version.  This could mean you would not be co-installable with projects you previously were (other projects that included opentracing as a dependency).  The only way to update would be to have the OTHER project update as well, everything being in lockstep.

yurishkuro: There's this code in the Tracer
```
        if self.active_span is not None and not ignore_active_span:
            parent = self.active_span
```

I assume `self.active_span` is coming from the base class, so it won't work without 2.x

mihau: @prometheanfire it's a fair point but making it compatible with both `2.x` and `1.x` will result in quite a bit of funky code with some `except ImportError` and `getattr` here and there, which makes for a less readable / maintainable code. Also the tests would need to be ran for both versions. @yurishkuro, @black-adder what do you think - is it worth the effort and consequences? I'm fine with forcing people to upgrade their other dependencies when they want to use a newer jaeger-client.

black-adder: For me, I'd prefer to just make a breaking change and release jaeger-client-python 4.0.0 along with some other long outstanding issues: https://github.com/jaegertracing/jaeger-client-python/issues/205

yurishkuro: +1, but "Remove tornado dependency #31" might be a tall order. Then again, without it  it's difficult to clean up the API so that ioloop is completely hidden.

tests/test_api.py
```diff
@@ -34,3 +34,6 @@ def tracer(self):
     def test_binary_propagation(self):
         # TODO binary codecs are not implemented at the moment
         pass
+
+    def is_parent(self, parent, span):
```
tovin07: Hi, I do not see this function is used any where (at least for now).

jaeger_client/tracer.py
```diff
@@ -84,6 +86,7 @@ def __init__(
         if tags:
             self.tags.update(tags)
 
+
```
tovin07: Is this new line inserted by accident?

jaeger_client/tracer.py
```diff
@@ -106,12 +109,18 @@ def __init__(
...
+        )
+
     def start_span(self,
                    operation_name=None,
                    child_of=None,
                    references=None,
                    tags=None,
-                   start_time=None):
+                   start_time=None,
+                   ignore_active_span=False):
```
tovin07:
```suggestion
                   ignore_active_span=False,
    ):
```

jaeger_client/tracer.py
```diff
@@ -184,6 +199,46 @@ def start_span(self,
...
         return span
 
+    def start_active_span(self,
+                          operation_name=None,
+                          child_of=None,
+                          references=None,
+                          tags=None,
+                          start_time=None,
+                          ignore_active_span=False,
+                          finish_on_close=True):
```
tovin07:
```suggestion
                          finish_on_close=True,
    ):
```

tests/test_api.py
mihau: This function is required by opentracing [here](https://github.com/opentracing/opentracing-python/blob/2ce13e2dd19e263daa5db88800cf6e84691a0b8e/opentracing/harness/api_check.py#L50).

jaeger_client/tracer.py
mihau: Yes, removed.

mihau:
I've spotted an issue with what I believe is a problem of `opentracing-python` tests, which led me to overriding a test in [this commit](https://github.com/jaegertracing/jaeger-client-python/pull/206/commits/2a2e8a3376d2185f0db4aec213442fc9a677d554).
I opened an issue in `opentracing-python` to figure out what to do with it.

jaeger_client/config.py
```diff
@@ -84,17 +84,20 @@ class Config(object):
...
     def __init__(self, config, metrics=None, service_name=None, metrics_factory=None,
-                 validate=False):
+                 validate=False, scope_manager=None):
         """
         :param metrics: an instance of Metrics class, or None. This parameter
             has been deprecated, please use metrics_factory instead.
         :param service_name: default service name.
             Can be overwritten by config['service_name'].
         :param metrics_factory: an instance of MetricsFactory class, or None.
+        :param scope_manager: a class implementing a scope manager, or None for
```
yurishkuro: why should this arg be a class rather than an actual object?

setup.py
```diff
@@ -40,7 +40,7 @@
         'threadloop>=1,<2',
         'thrift',
         'tornado>=4.3,<5',
-        'opentracing>=1.2.2,<2',
+        'opentracing>=2.0,<3.0',
```
yurishkuro: pleasantly surprised that crossdock test works with `'opentracing_instrumentation>=2,<3'`

jaeger_client/config.py
mihau: for no good reason it seems, looking at it now I'd say that passing an object would probably better since it's more flexible. I'll try it out and see if it doesn't break anything.

yurishkuro: thanks. Passing object makes more sense to me since it's supposed to be a singleton, so there's little reason why the tracer should be instantiating it from a class.

yurishkuro:
Aside from the scope manager passed as class vs. instance, this lgtm otherwise (i.e. WIP can be removed).

mihau:
@yurishkuro and how do you feel about this [commit](https://github.com/jaegertracing/jaeger-client-python/pull/206/commits/2a2e8a3376d2185f0db4aec213442fc9a677d554) do we leave it, or do we get it fixed in `python-opentracing` first?

yurishkuro:
I think it's ok to leave it. Is there a ticket for opentracing-python? We should added it to the comments under `// TODO remove after {ticket} is fixed

jaeger_client/config.py
mihau: Makes sense, the tests pass with this change.

mihau:
Ok, cool. I've added the TODO note. Removing the WIP status.

jaeger_client/tracer.py
```diff
@@ -129,10 +137,16 @@ def start_span(self,
...
         :param start_time: an explicit Span start time as a unix timestamp per
             time.time()
+        :param ignore_active_span: an explicit flag that ignores the current
+            active :class:`Scope` and creates a root :class:`Span`
 
         :return: Returns an already-started Span instance.
         """
         parent = child_of
+
+        if self.active_span is not None and not ignore_active_span:
```
yurishkuro: I think it also needs `and not parent`, because if child_of is passed explicitly, the active span should not be inherited (or am I mistaken?)

jaeger_client/tracer.py
```diff
@@ -189,6 +203,46 @@ def start_span(self,
...
+
+        :return: a :class:`Scope`, already registered via the :class:`ScopeManager`.
+        """
+        span = self.start_span(
+            operation_name,
+            child_of,
+            references,
+            tags,
+            start_time,
+            ignore_active_span,
```
yurishkuro: please use named arguments

tests/test_api.py
```diff
@@ -34,3 +34,22 @@ def tracer(self):
...
+        return span.parent_id == getattr(parent, 'span_id', None)
+
+    # NOTE: this overrides a method defined in ApiComaptibilityCheckMixin in
+    # order to add the scope.close() call, which prevents the tracer from
+    # leaving an active scope in thread local storage
+    # TODO: remove after
+    # https://github.com/opentracing/opentracing-python/issues/117 is fixed
```
yurishkuro: it looks like the code is already fixed, but we just need a new release

mihau: You're right, any idea when is the `2.1.0` coming? If it's not coming anytime soon then I'll leave the override and change the `TODO` "remove after moving to 2.1.0 of python-opentracing"

yurishkuro: I'll try to get a release in the next few days. Meanwhile a better ticket to reference is https://github.com/opentracing/opentracing-python/issues/119

jaeger_client/tracer.py
mihau: I think you're right, it makes sense. Although one could use `ignore_active_span` together with `child_of` to pass it explicitly. Not exactly intuitive but still an option. The tests don't seem to enforce it in any way (maybe it would be appropriate to add one to do so?).
I added the condition, you might find the formatting controversial (multiline `if`), let me know if you prefer a different way.

jaeger_client/tracer.py
mihau: done

tests/test_api.py
mihau: done

PR #207, opened 2018-08-20
Support `jaeger-baggage` header for ad-hoc baggage

yurishkuro:

### DIFF
--- a/jaeger_client/codecs.py
+++ b/jaeger_client/codecs.py
@@ -19,6 +19,7 @@ from opentracing import (
     SpanContextCorruptedException,
 )
 from .constants import (
+    BAGGAGE_HEADER_KEY,
     BAGGAGE_HEADER_PREFIX,
     DEBUG_ID_HEADER_KEY,
     TRACE_ID_HEADER,
@@ -43,11 +44,13 @@ class TextCodec(Codec):
                  url_encoding=False,
                  trace_id_header=TRACE_ID_HEADER,
                  baggage_header_prefix=BAGGAGE_HEADER_PREFIX,
-                 debug_id_header=DEBUG_ID_HEADER_KEY):
+                 debug_id_header=DEBUG_ID_HEADER_KEY,
+                 baggage_header=BAGGAGE_HEADER_KEY):
         self.url_encoding = url_encoding
         self.trace_id_header = trace_id_header.lower().replace('_', '-')
         self.baggage_prefix = baggage_header_prefix.lower().replace('_', '-')
         self.debug_id_header = debug_id_header.lower().replace('_', '-')
+        self.baggage_header = baggage_header
         self.prefix_length = len(baggage_header_prefix)
 
     def inject(self, span_context, carrier):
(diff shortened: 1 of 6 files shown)


jaeger_client/codecs.py
```diff
@@ -110,15 +113,27 @@ def extract(self, carrier):
                 if self.url_encoding:
                     value = urllib_parse.unquote(value)
                 debug_id = value
-        if not trace_id and baggage:
-            raise SpanContextCorruptedException('baggage without trace ctx')
```
yurishkuro: I could not find any rationalization for why this exception had to be thrown in this scenario.

tests/test_codecs.py
```diff
@@ -208,19 +208,6 @@ def test_context_from_readable_headers(self):
                 'hermes': 'LaBarbara Hermes',
             }
 
-    def test_baggage_without_trace_id(self):
```
yurishkuro: moved down (outside of class) to be able to get `tracer` fixture. Not sure if the same possible when subclassing `TestCodecs(unittest.TestCase)`.

tests/test_thrift.py
```diff
@@ -69,10 +70,6 @@ def serialize(span_id):
         span.finish()
         _marshall_span(span)
 
-    trace_id = 0
```
yurishkuro: how was this testing "large trace IDs"? 0 wasn't even a valid id.

jaeger_client/tracer.py
```diff
@@ -157,10 +157,12 @@ def start_span(self,
                     tags = tags or {}
                     for k, v in six.iteritems(sampler_tags):
                         tags[k] = v
-            elif self.is_debug_allowed(operation_name):  # have debug id
+            elif parent.debug_id and self.is_debug_allowed(operation_name):
```
isaachier: I think this conflates `debug_id` with debug flag. A debug ID is a specific header along the lines of `jaeger-debug-id: foo`, which then associates an ID `"foo"` with a new span. That span also sets its debug flag to true (assuming throttling allows it). In retrospect, maybe we should revert this change.

jaeger_client/tracer.py
```diff
@@ -157,10 +157,12 @@ def start_span(self,
...
                 flags = SAMPLED_FLAG | DEBUG_FLAG
                 tags = tags or {}
                 tags[self.debug_id_header] = parent.debug_id
+            if parent and parent.baggage:
+                baggage = dict(parent.baggage)  # TODO do we need to clone?
```
isaachier: If the `dict` is immutable, you should not have to clone. Immutability is definitely the approach in other clients, IDK about this one.

jaeger_client/tracer.py
yurishkuro: where do you se a disconnect? The only way we get to this part is if we have no parent OR the parent is empty but has a debug_id tag set. Previously those were the only condition, now you can also get here with the parent empty (as in: no trace id), but having the baggage. Yet it may not have the debug_id tag. Previously this `elif` part always implied there are the debug_id tag, now we need to check for it.

isaachier: OK sorry my confusion here was based on a misunderstanding here regarding the necessity of a parent. I had thought you were implying that all debug flags must be propagated from a real span object, not just a header. I forgot the header is transformed into an empty span context with a debug ID, but nothing else.

PR #208, opened 2018-08-27
Initial Sender implementation

rmfitzpatrick:
Continuing on with the adoption of Sender from https://github.com/jaegertracing/jaeger-client-python/pull/186 and addresses requested changes.

~~This addition still has Reporter manage the consume_queue background callback as it possesses the ErrorReporter and ReporterMetrics attributes (https://github.com/jaegertracing/jaeger-client-python/pull/186#pullrequestreview-144048686).  I plan on moving that functionality to Sender while incorporating appropriate UDP batching in a subsequent PR before adding an HTTPSender.~~

~~edit:  Given https://github.com/jaegertracing/jaeger-client-python/pull/208#issuecomment-416274013 I will add UDP batching to this PR with the expectation that the Reporter span queue management is to roughly stay as is functionally.~~

~~edit 2: UDP batching has been added.~~ https://github.com/jaegertracing/jaeger-client-python/pull/208#issuecomment-430298109  -- Just addressing breaking changes and will introduce UDP batching in another PR.

### DIFF
--- a/jaeger_client/reporter.py
+++ b/jaeger_client/reporter.py
@@ -24,6 +24,7 @@ from tornado.concurrent import Future
 from .constants import DEFAULT_FLUSH_INTERVAL
 from . import ioloop_util
 from .metrics import Metrics, LegacyMetricsFactory
+from jaeger_client.thrift_gen.agent import Agent
 from .senders import UDPSender
 from .utils import ErrorReporter
 
@@ -97,7 +98,13 @@ class Reporter(NullReporter):
         from threading import Lock
 
         # TODO for next major rev: remove channel param in favor of sender
-        self._sender = sender or self._create_default_sender(channel)
+        self.agent = Agent.Client(channel, self)
+        self._sender = sender or UDPSender(
+            port=channel._reporting_port,
+            host=channel._host,
+            io_loop=channel.io_loop,
+            agent=self.agent
+        )
         self.queue_capacity = queue_capacity
         self.batch_size = batch_size
         self.metrics_factory = metrics_factory or LegacyMetricsFactory(metrics or Metrics())
@@ -108,36 +115,24 @@ class Reporter(NullReporter):
         if queue_capacity < batch_size:
             raise ValueError('Queue capacity cannot be less than batch size')
 
-        self.io_loop = io_loop or self.fetch_io_loop(channel, self._sender)
-
-        if self.io_loop is None:
-            self.logger.error('Jaeger Reporter has no IOLoop')
+        if io_loop:
+            self.io_loop = io_loop
         else:
-            self.queue = tornado.queues.Queue(maxsize=queue_capacity)
-            self.stop = object()
-            self.stopped = False
-            self.stop_lock = Lock()
-            self.flush_interval = flush_interval or None
-            self.io_loop.spawn_callback(self._consume_queue)
+            self.io_loop = self._sender._io_loop
 
-    @staticmethod
-    def fetch_io_loop(channel, sender):
-        if channel:
-            return channel.io_loop
-        elif sender:
-            return sender.io_loop
-        return None
-
-    def _create_default_sender(self, channel):
-        sender = UDPSender(
-            port=channel.reporting_port,
-            host=channel.host,
-            io_loop=channel.io_loop
-        )
-        return sender
+        self.queue = tornado.queues.Queue(maxsize=queue_capacity)
+        self.stop = object()
+        self.stopped = False
+        self.stop_lock = Lock()
+        self.flush_interval = flush_interval or None
+        self.io_loop.spawn_callback(self._consume_queue)
 
     def set_process(self, service_name, tags, max_length):
         self._sender.set_process(service_name, tags, max_length)
+
+    # method for protocol factory
+    def getProtocol(self, transport):
+        return self._sender.getProtocol(transport)
 
     def report_span(self, span):
         # We should not be calling `queue.put_nowait()` from random threads,
--- a/jaeger_client/reporter.py
+++ b/jaeger_client/reporter.py
@@ -26,10 +26,9 @@ from .constants import DEFAULT_FLUSH_INTERVAL
 from . import thrift
 from . import ioloop_util
 from .metrics import Metrics, LegacyMetricsFactory
+from .senders import UDPSender
 from .utils import ErrorReporter
 
-from thrift.protocol import TCompactProtocol
-from jaeger_client.thrift_gen.agent import Agent
 
 default_logger = logging.getLogger('jaeger_tracing')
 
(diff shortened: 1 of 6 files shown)


yurishkuro:
One meta-comment I have: I think we will keep running into issues with backwards-compatibility of refactoring as long as the Reporter is using IOLoop for its internal queue mgmt. If we re-implement it with a separate thread (the way Lightstep did) and remove the dependency on IOLoop, the rest of the refactoring will become much easier.

rmfitzpatrick:
@yurishkuro, in the wake of https://github.com/jaegertracing/jaeger-client-python/issues/205 it seems like it could be a good time to address the tornado-ectomy in a subsequent PR, but I'm not 100% of the proposed architecture.  Is the Reporter intended to maintain ownership of the loop or should all of that be delegated to its Sender while it just maintains the basic NullReporter interface?  I'm getting the (potentially incorrect) impression that you and @black-adder are approaching the solution differently.

Happy to move this discussion to https://github.com/jaegertracing/jaeger-client-python/issues/31.

yurishkuro:
I think it's best to keep the same architecture as in other Jaeger clients: 
  * the Reporter runs a background thread and a queue
  * queue is used to shed load if the sender is not fast enough
  * senders do not need to be multi-threaded, only Reporter is

rmfitzpatrick:
Basic UDP batching has been added.  I used jaeger-client-node as a rough template but am unclear if my thrift usage is optimal to get the content length as it requires serialization before properly encoding the batch.

rmfitzpatrick:
@black-adder, wanted to make sure you knew this is where I respond to your https://github.com/jaegertracing/jaeger-client-python/pull/186 feedback.  Thanks for your thorough review.

rmfitzpatrick:
@black-adder & @yurishkuro, is this still on the block as a precursor to an HTTPSender?  I don't believe it introduces any breaking changes so both senders shouldn't necessitate a major version (re: https://github.com/jaegertracing/jaeger-client-python/issues/205).

yurishkuro:
@rmfitzpatrick this PR is very large. I saw you added UDP batching, is it possible to pull this out into an independent PR and merge separately from HTTP refactoring?

rmfitzpatrick:
@yurishkuro, went ahead and removed that commit.

yurishkuro:
@tiffon you said you can review this?

jaeger_client/reporter.py
```diff
@@ -120,14 +122,29 @@ def __init__(self, channel, queue_capacity=100, batch_size=10,
             self.flush_interval = flush_interval or None
             self.io_loop.spawn_callback(self._consume_queue)
 
-        self._process_lock = Lock()
-        self._process = None
+    @staticmethod
+    def _fetch_io_loop(channel, sender):
```
tiffon: Quick question: `_fetch_io_loop` is used once. What's the advantage to using a static method vs inlining?

jaeger_client/reporter.py
```diff
@@ -120,14 +122,29 @@ def __init__(self, channel, queue_capacity=100, batch_size=10,
...
+        if channel:
+            return channel.io_loop
+        elif sender:
+            return sender.io_loop
+        return None
+
+    def _create_default_sender(self, channel, agent):
```
tiffon: Since `_create_default_sender` is used in only one place why not inline it?

jaeger_client/reporter.py
rmfitzpatrick: This was introduced in https://github.com/jaegertracing/jaeger-client-python/pull/186/commits/435b5329bb0fb605342c9dd4542064522d2d696c, and from the looks of it is was for readability and testability.  Considering the io loop should come from the sender, I went ahead and corrected inline.

jaeger_client/reporter.py
rmfitzpatrick: Made inline.

rmfitzpatrick:
@tiffon, any chance you've had a moment to look over my updates?

jaeger_client/senders.py
```diff
@@ -0,0 +1,137 @@
...
+
+class Sender(object):
+    def __init__(self, io_loop=None):
+        from threading import Lock
+        self._io_loop = io_loop or self._create_new_thread_loop()
+        self._process_lock = Lock()
+        self._process = None
+        self.spans = []
+
+    def append(self, span):
```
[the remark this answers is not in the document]
rmfitzpatrick: Sorry for the belatedness of my reply, but I have some cycles to spend on this now.  I'm definitely up for adopting the proposed api, but I don't think I understand what you're referring to regarding the other jaeger clients.

jaeger-client-node: the RemoteReporter maintains a flush interval: https://github.com/jaegertracing/jaeger-client-node/blob/master/src/reporters/remote_reporter.js#L37

jaeger-client-java: the RemoteReporter maintains a flush timer: https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/internal/reporters/RemoteReporter.java#L65

Jaeger-client-go: the remoteReporter.processQueue() follows similar behavior: https://github.com/jaegertracing/jaeger-client-go/blob/master/reporter.go#L252

Full disclosure that I haven't poured over these and could be missing something in each case.

jaeger_client/senders.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Create a daemonized thread that will run Tornado IOLoop.
+        :return: the IOLoop backed by the new thread.
+        """
+        self._thread_loop = ThreadLoop()
+        if not self._thread_loop.is_ready():
+            self._thread_loop.start()
+        return self._thread_loop._io_loop
+
+    def getProtocol(self, transport):
```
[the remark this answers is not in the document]
rmfitzpatrick: Yes, this seems to be a required thrift method.  This is currently a method in [the reporter](https://github.com/jaegertracing/jaeger-client-python/commit/d98cae40c0b0e702b0fbb826af1cbe53fa553347#diff-8a53213519652bcef509b8fbddbdfe86R212), and was moved to the udp sender as the protocol may depend on the sender type.  https://github.com/jaegertracing/jaeger-client-python/pull/186/commits/435b5329bb0fb605342c9dd4542064522d2d696c#r208288056 observed it was a breaking change, so is now wrapped by the reporter to retrieve from the sender as the protocol factory could vary for each sender type.  An http sender I'm currently uses the binary protocol for example.

jaeger_client/senders.py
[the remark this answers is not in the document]
rmfitzpatrick: Also, it is possible, but expense to get the exact byte count of a thrift message and I'll be sure to propose proper udp batching in a subsequent PR (was removed from this one due to the added complexity).

shuaichang:
Just curious if this effort is still continued?  We have use cases in Serverless functions to trace with Jaeger  and Python.

rmfitzpatrick:
@shuaichang I've unfortunately let this slip on my todo list, since I think it and http functionality are unlikely to land until https://github.com/jaegertracing/jaeger/issues/1269 is sorted out (which it may have been but I've lost track).  fwiw I'm using a fork of this library that removes internal tornado usage and supports http submissions: https://github.com/signalfx/jaeger-client-python/blob/sfx-release/jaeger_client/config.py#L137.

I would like this to continue but may need someone to take the reins if unified client design is still tbd.

PR #209, opened 2018-09-03
Mark reporter_queue_size Config option as valid

TimDumol:
If you add reporter_queue_size as an option in a jaeger_client.Config,
it currently raises an error, even if it's a valid option if
validate=True.

### DIFF
--- a/jaeger_client/config.py
+++ b/jaeger_client/config.py
@@ -116,6 +116,7 @@ class Config(object):
                         'tags',
                         'enabled',
                         'reporter_batch_size',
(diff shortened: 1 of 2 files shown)


yurishkuro:
do you mind adding a unit test? the test should be breaking before your main change.

yurishkuro:
"item in the list" is an implementation detail. Unit tests checks external functional behavior that you discovered wasn't working.

I restarted the build, looks like some Travis issue.

yurishkuro:
Thank you!

Issue #212, opened 2018-09-29
set_tag always converts values to string


trondhindenes:
## Requirement - what kind of business use case are you trying to solve?
Trying to post spans containing tags with int and/or boolean values

## Problem - what in Jaeger blocks you from solving the requirement?
The `set_tag` methods seems to convert all incoming values to stringed values.

PR #213, opened 2018-09-29
Preserve datatype of tag values

trondhindenes:
This PR implements type checking on the `set_tag` so that kv pairs where the value is not a string gets stored using the correct thrift plumbing. It implements support for booleans, integers (long), doubles in addition to the existing string type, which is also the fallback type.

Here's an example of using it:
screenshot from 2018-09-29 22-02-28

Fixes https://github.com/jaegertracing/jaeger-client-python/issues/212

### DIFF
--- a/jaeger_client/span.py
+++ b/jaeger_client/span.py
@@ -81,11 +81,10 @@ class Span(opentracing.Span):
             if key == ext_tags.SAMPLING_PRIORITY and not self._set_sampling_priority(value):
                 return self
             if self.is_sampled():
-                tag = thrift.make_string_tag(
+                tag = thrift.make_tag(
                     key=key,
                     value=value,
-                    max_length=self.tracer.max_tag_value_length,
-                )
+                    max_length=self.tracer.max_tag_value_length, )
                 self.tags.append(tag)
         return self
 
--- a/setup.py
+++ b/setup.py
@@ -54,9 +54,9 @@ setup(
             'mock==1.0.1',
             'pycurl>=7.43,<8',
             'pytest>=3.6.0',
-            'pytest-cov',
+            'pytest-cov==2.5.1',
             'coverage<4.4',  # can remove after https://bitbucket.org/ned/coveragepy/issues/581/44b1-44-breaking-in-ci
-            'pytest-timeout',
+            'pytest-timeout==1.3.1',
             'pytest-tornado',
             'pytest-benchmark[histogram]>=3.0.0rc1',
             'pytest-localserver',
@@ -65,7 +65,7 @@ setup(
             'codecov',
             'tchannel>=0.27', # This is only used in python 2
             'opentracing_instrumentation>=2,<3',
-            'prometheus_client',
+            'prometheus_client==0.3.1',
         ]
     },
 )
--- a/jaeger_client/thrift.py
+++ b/jaeger_client/thrift.py
@@ -47,7 +47,31 @@ def _to_string(s):
         return str(e)
 
 
-def make_string_tag(key, value, max_length):
+def make_tag(key, value, max_length):
+    if type(value).__name__ == 'bool':  # isinstance doesnt work on booleans
+        return _make_bool_tag(
+            key=key,
+            value=value
+        )
+    elif isinstance(value, int):
+        return _make_long_tag(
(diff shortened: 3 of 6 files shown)


trondhindenes:
Not sure why builds are failing, looks like that might be a problem outside of this PR...

jaeger_client/span.py
```diff
@@ -81,11 +81,27 @@ def set_tag(self, key, value):
             if key == ext_tags.SAMPLING_PRIORITY and not self._set_sampling_priority(value):
                 return self
             if self.is_sampled():
-                tag = thrift.make_string_tag(
-                    key=key,
-                    value=value,
-                    max_length=self.tracer.max_tag_value_length,
-                )
```
yurishkuro: Thanks for the PR.

Nit: I think this is not the only place. I would rather create a single method thrift.create_tag and do the dispatching by type inside it, to new private methods.

trondhindenes: sure I can try and do that

trondhindenes:
I'm noticing that this change has some implications for an existing test:
```
@pytest.mark.parametrize('span_type,expected_tags', [
    ('root', {
        'sampler.type': 'const',
        'sampler.param': 'True',
    }),
```
Since the test expects the string `'True'` the test `test_tracer_tags_on_root_span` is now failing. Not quite sure if its okay to just adjust the test params or if this is a sign of trouble ahead.

Peeking over in the Java implementation of Opentracing, I'm finding this:
`assertEquals(true, tags.get("sampler.param"));`
which should indicate that the correct (or at least valid) type for the `sampler.param` tag is a boolean, so I guess it's okay to adjust the failing test? I'd really appreciate any input on this.

trondhindenes:
did a little refactoring of the find_tag method to make it more type-aware, not sure if that's the best way to do it.

trondhindenes:
Will squash. As for passing the build, looks like there's a coverage module missing. Happy to add it, had to do that manually on my computer aswell.

trondhindenes:
I've checked this a bit further.
locking `pytest-cov==2.5.1` will remove the version conflict for the `coverage` package.
As for the two failing prometheus tests, those are failing for me on the master branch as well. As far as I can see the last successful master build was a while ago, and it's difficult to see the exact package versions used.
However, by locking the versions to the versions they were at the time of the last build (especially prometheus-client which has been bumped from 0.3.1 to 0.4.0) the builds are succeeding. 

I realize that these changes are outside of the domain of my little pull request, so I fully expect this PR to not get accepted as it standard right now. Still pushing it to at least have a non-failing build. Open to suggestions on what to do here.

trondhindenes:
Hm still failing. I have honestly no idea what's failing. We're probably just gonna run this PR as-is from our internal package feed to get the functionality we need. Let me know if there's anything more I (realistically) can do to get this merged in here.

yurishkuro:
the other builds seem to be failing with the same

jaeger_client/codecs.py
```diff
@@ -110,15 +113,27 @@ def extract(self, carrier):
...
+            # reset all IDs
+            trace_id, span_id, parent_id, flags = None, None, None, None
+        if not trace_id and not debug_id and not baggage:
             return None
         return SpanContext(trace_id=trace_id, span_id=span_id,
                            parent_id=parent_id, flags=flags,
-                           baggage=baggage)
+                           baggage=baggage, debug_id=debug_id)
+
+    def _parse_baggage_header(self, header, baggage):
```
yurishkuro: where did this change come from? I thought it wasn't part of the earlier PR.

setup.py
```diff
@@ -65,7 +65,7 @@
             'codecov',
             'tchannel>=0.27', # This is only used in python 2
             'opentracing_instrumentation>=2,<3',
-            'prometheus_client',
+            'prometheus_client==0.3.1',
```
yurishkuro: Thanks for pinning. We can revisit that separately.

trondhindenes:
Sorry for the mess. The PR should be in good shape now. Added a few new tests for the new thrift methods as well.

yurishkuro:
PS: consider adding yourself to https://github.com/jaegertracing/jaeger/issues/207

PR #214, opened 2018-10-03
Update check-license.sh

NavneetKatiyar:


yurishkuro:
there's no change here

PR #215, opened 2018-10-26
Fix handling of missing headers in the b3 codec

cshowe:
The b3 codec uses .get() to extract headers and so does not throw a key
error if headers are missing.  However many of these headers are decoded
via "header_to_hex" which throws an exception on a None value.  Other
codecs handle the missing header case by returning None, update the b3
codec to follow this convention.

Signed-off-by: Caleb Howe <Caleb.S.Howe@gmail.com>

### DIFF
--- a/jaeger_client/codecs.py
+++ b/jaeger_client/codecs.py
@@ -295,7 +295,6 @@ class B3Codec(Codec):
         elif span_context.flags & SAMPLED_FLAG == SAMPLED_FLAG:
             carrier[self.sampled_header] = '1'
 
-
     @staticmethod
     def _get_header_value_hex(carrier, header_lc_to_header_orig, header_lc):
         header = header_lc_to_header_orig.get(header_lc)
--- a/jaeger_client/codecs.py
+++ b/jaeger_client/codecs.py
@@ -295,32 +295,27 @@ class B3Codec(Codec):
         elif span_context.flags & SAMPLED_FLAG == SAMPLED_FLAG:
             carrier[self.sampled_header] = '1'
 
-    @staticmethod
-    def _get_header_value_hex(carrier, header_lc_to_header_orig, header_lc):
-        header = header_lc_to_header_orig.get(header_lc)
-        if header is None:
-            return None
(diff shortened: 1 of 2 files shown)


jaeger_client/codecs.py
```diff
@@ -295,15 +295,25 @@ def inject(self, span_context, carrier):
...
+            return None
+        header_value = carrier.get(header)
+        if header_value is not None:
+            header_value = header_to_hex(header_value)
+        return header_value
+
     def extract(self, carrier):
         if not isinstance(carrier, dict):
             raise InvalidCarrierException('carrier not a dictionary')
         lowercase_keys = dict([(k.lower(), k) for k in carrier])
```
yurishkuro: I wonder if it would be simpler and more efficient to just loop through the carrier keys, e.g.
```
for key, value in six.itemitems(carrier):
    if value is None:
        continue  # is this even possible in Python?
    lowerKey = key.lower()
    if lowerKey == self._trace_header_lc:
        # parse
    elif lowerKey == self._span_header_lc:
        # parse
```
This avoids creating an intermediate dictionary, and partially addresses your issue

yurishkuro:
@cshowe ping - are you still interested in this PR?

cshowe:
Yes, just fell off my radar.  I should address your comments today.

tests/test_codecs.py
```diff
@@ -339,6 +339,20 @@ def test_b3_extract(self):
         span_context = codec.extract(carrier)
         assert span_context.flags == 0x01
 
+        # validate present debug header with falsy value
+        carrier = {'X-b3-SpanId': 'a2fb4a1d1a96d312', 'X-B3-flags': '0',
+                   'X-B3-traceId': '463ac35c9f6413ad48485a3953bb6124'}
```
yurishkuro: interesting...
```python
>>> header = '463ac35c9f6413ad48485a3953bb6124'
>>> int(header, 16)
93351075330931896558786731617803788580L
```
perhaps supporting 128bit IDs (#228) is not going to be particularly difficult.

PR #216, opened 2018-10-30
Adding delete baggage item API for span

Fantoccini:


jaeger_client/span.py
```diff
@@ -142,6 +142,22 @@ def set_baggage_item(self, key, value):
     def get_baggage_item(self, key):
         return self.context.baggage.get(key)
 
+    def remove_baggage_item(self, key):
```
yurishkuro: the API of the Span class is dictated by the OpenTracing API, we cannot arbitrarily add new methods. I believe if we simply implement the change to the SpanContext below, then you can achieve the same effect of removing baggage item via `span.set_baggage_item(key, None)`.

yurishkuro:
please make sure all commits are signed (see CONTRIBUTING file).

tests/test_span.py
```diff
@@ -40,6 +40,9 @@ def test_baggage():
     span.set_baggage_item('X_y', '123')
     assert span.get_baggage_item('X_y') == '123'
     assert span.get_baggage_item('x-Y') is None
+    span.set_baggage_item('x-Y', None).set_baggage_item('z', None)
```
yurishkuro: Why is the first one `x-Y`? It is already None. I am not sure what you're trying to test with that.

I suggest adding something like `set_baggage_item('nonExistingKey', None)` to test that it won't cause an error.

tests/test_span.py
```diff
@@ -40,6 +40,9 @@ def test_baggage():
...
+    assert span.get_baggage_item('X_y') == '123'
+    assert span.get_baggage_item('z') is None
```
yurishkuro: this does not really test that the key was removed from the baggage

tests/test_span_context.py
```diff
@@ -33,6 +34,14 @@ def test_with_baggage_items():
     assert ctx1.baggage != ctx2.baggage
     baggage1['a'] = 'b'
     assert ctx1.baggage == ctx2.baggage
+    ctx3 = ctx2.with_baggage_item('a', None)
+    assert ctx2.trace_id == ctx3.trace_id
+    assert ctx2.span_id == ctx3.span_id
+    assert ctx2.parent_id == ctx3.parent_id
+    assert ctx2.flags == ctx3.flags
```
yurishkuro: the 4 asserts above are not necessary

tests/test_span_context.py
```diff
@@ -33,6 +34,14 @@ def test_with_baggage_items():
     assert ctx1.baggage != ctx2.baggage
     baggage1['a'] = 'b'
     assert ctx1.baggage == ctx2.baggage
+    ctx3 = ctx2.with_baggage_item('a', None)
```
yurishkuro: please add blank link above

Fantoccini:
Thanks @yurishkuro

Fantoccini:
Hi @yurishkuro , can you please help me to restart the travis build? The coverage job somehow stuck there. Thanks.

yurishkuro:
restarted

PR #218, opened 2018-11-14
Allow setting of the Jaeger Agent Host via environment variable.

timstoop:
Hi,

On Kubernetes, it would be very convenient if the Jaeger Agent Host can be set via an environment variable. This is also the method described in the Jaeger for Kubernetes docs (https://github.com/jaegertracing/jaeger-kubernetes). Adding this PR will make sure that those Jaeger on Kubernetes docs also work out of the box with projects created using this library. I added a test to make sure it's working as expected.

Please let me know if you need anything changed in here!

### DIFF
--- a/jaeger_client/config.py
+++ b/jaeger_client/config.py
@@ -259,6 +259,11 @@ class Config(object):
         try:
             return self.local_agent_group()['reporting_host']
         except:
+            pass
+
+        if os.getenv('JAEGER_AGENT_HOST') is not None:
+            return os.getenv('JAEGER_AGENT_HOST')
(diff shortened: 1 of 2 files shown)


timstoop:
I have no idea why the crossdock build fails on Travis?

yurishkuro:
I restarted the build on master, to see if this is related to the PR or not.

https://travis-ci.org/jaegertracing/jaeger-client-python/jobs/449920101

timstoop:
Seems to be failing again, so going to assume it's not my PR (unless you tell me differently!).

PR #219, opened 2018-11-19
Wrap parsing of throttling response in try block to avoid errors

isaachier:
Signed-off-by: Isaac Hier <ihier@uber.com>

(internal ref T2245255)


isaachier:
Have these tests been failing until recently?

yurishkuro:
the crossdock seems to have a hard time since recently, but the unit tests failures I haven't seen before, they look more like linter issues, probably because of a new linter release (we should really be pinning those tools)

yurishkuro:
trying to fix crossdock tests in #222

isaachier:
Looks good now. Awaiting approval.

PR #220, opened 2018-11-20
Remove pytest-tornado from dependencies

isaachier:
pytest-tornado is no longer maintained. A pull request was submitted months ago for an error we have seen repeatedly in our unit tests (https://github.com/eugeniy/pytest-tornado/pull/38). Nonetheless, the pull request has not been merged. We should just use regular tornado testing utilities to avoid this unmaintained package.

### DIFF
--- a/tests/test_crossdock.py
+++ b/tests/test_crossdock.py
@@ -24,143 +24,126 @@ from mock import MagicMock
 if six.PY2:
     from crossdock.server import server
 from tornado.httpclient import HTTPRequest
+import tornado.testing
 from jaeger_client import Tracer, ConstSampler
 from jaeger_client.reporter import InMemoryReporter
 from crossdock.server.endtoend import EndToEndHandler, _determine_host_port, _parse_host_port
 
 tchannel_port = "9999"
-
-
-@pytest.fixture
-def app():
-    """Required by pytest-tornado's http_server fixture"""
-    s = server.Server(int(tchannel_port))
-    s.tchannel.listen()
-    return server.make_app(s, EndToEndHandler())
-
-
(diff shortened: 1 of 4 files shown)


tests/test_crossdock.py
```diff
@@ -24,143 +24,126 @@
...
-def app():
-    """Required by pytest-tornado's http_server fixture"""
-    s = server.Server(int(tchannel_port))
-    s.tchannel.listen()
-    return server.make_app(s, EndToEndHandler())
-
-
-# noinspection PyShadowingNames
-@pytest.yield_fixture
-def tracer():
```
yurishkuro: why do you need to change this? it's not dependent on pytest-tornado. If you simply call get_tracer() from the unit test, then you're never closing the tracer.

isaachier: I tried it. Using any generators within a test case without marking it as `@tornado.testing.gen_test` causes an error. If I add that, it won't work at all because it is not meant for object methods, only free functions.

yurishkuro: does `tornado.testing` not work with free functions at all? I have that class extension pattern, makes a lot of testing annotations unusable.

isaachier: I totally agree about that. On the other hand, I saw no way around this given that pytest-tornado hasn't been maintained in a while and pytest has deprecated some key features.

yurishkuro:
I pinned pytest in #221 so that it does not produce Py4 warnings. Defer the problem to another day - I'd rather we get rid of tornado dependency altogether (e.g. #31)

isaachier:
Cool.

PR #230, opened 2019-01-04
 Add 128bit trace_id support

### LINKED ISSUE #228: Support for 128 bit trace_id
I'm trying to implement Jaeger tracing in a flask app which will be running on [Fission](https://fission.io/) 

The Jaeger Python client currently does not support 128bit tracing id while some of the other clients do support it.

If the support of 128 bit trace_id is planned for Python client I would like to make the change and send PR for it.

bhavin192:
- Creates a new property in Config `generate_128bit_trace_id` if set
  to `True` then 128bit trace_ids are generated
  - Adds support for environment variable `JAEGER_TRACEID_128BIT`. If
  the value is 'true' then 128bit trace_ids are generated
- Tracer takes variable `generate_128bit_trace_id`, if the value is `True`
  then `max_trace_id_bits` is set to `constants._max_trace_id_bits`
  - Adds new function `_random_id(bitsize)`
  - Deprecate `random_id()` in favor of `_random_id(bitsize)`
- Deprecate `MAX_ID_BITS` in favor of `_max_trace_id_bits` and `_max_id_bits`
- Set low and high bits of traceId correctly
  - Uses `_id_to_high()` and `_id_to_low()` to do that

Fixes #228 
ref: https://github.com/jaegertracing/jaeger/issues/858

### DIFF
--- a/jaeger_client/thrift.py
+++ b/jaeger_client/thrift.py
@@ -26,6 +26,24 @@ if six.PY3:
     long = int
 
 
+def _id_to_low(big_id):
+    """
+    :param big_id: id in integer
+    :return: Returns the right most 64 bits of big_id
+    """
+    if big_id is not None:
+        return big_id & (_max_unsigned_id - 1)
+
+
+def _id_to_high(big_id):
+    """
+    :param big_id: id in integer
+    :return: Returns the left most 64 bits of big_id
+    """
+    if big_id is not None:
+        return (big_id >> 64) & (_max_unsigned_id - 1)
+
+
 def id_to_int(big_id):
     if big_id is None:
         return None
@@ -150,8 +168,8 @@ def make_jaeger_batch(spans, process):
     for span in spans:
         with span.update_lock:
             jaeger_span = ttypes.Span(
-                traceIdLow=id_to_int(span.trace_id),
-                traceIdHigh=0,
+                traceIdLow=id_to_int(_id_to_low(span.trace_id)),
+                traceIdHigh=id_to_int(_id_to_high(span.trace_id)),
                 spanId=id_to_int(span.span_id),
                 parentSpanId=id_to_int(span.parent_id) or 0,
                 operationName=span.operation_name,
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -44,6 +44,7 @@ class Tracer(opentracing.Tracer):
         self, service_name, reporter, sampler, metrics=None,
         metrics_factory=None,
         trace_id_header=constants.TRACE_ID_HEADER,
+        generate_128bit_trace_id=False,
         baggage_header_prefix=constants.BAGGAGE_HEADER_PREFIX,
         debug_id_header=constants.DEBUG_ID_HEADER_KEY,
         one_span_per_rpc=False, extra_codecs=None,
@@ -60,6 +61,8 @@ class Tracer(opentracing.Tracer):
         self.debug_id_header = debug_id_header
         self.one_span_per_rpc = one_span_per_rpc
         self.max_tag_value_length = max_tag_value_length
+        self.max_trace_id_bits = constants._max_trace_id_bits if generate_128bit_trace_id \
+            else constants._max_id_bits
         self.codecs = {
             Format.TEXT_MAP: TextCodec(
                 url_encoding=False,
@@ -144,8 +147,8 @@ class Tracer(opentracing.Tracer):
             tags.get(ext_tags.SPAN_KIND) == ext_tags.SPAN_KIND_RPC_SERVER
 
         if parent is None or not parent.has_trace:
-            trace_id = self.random_id()
-            span_id = trace_id
+            trace_id = self._random_id(self.max_trace_id_bits)
+            span_id = self._random_id(constants._max_id_bits)
             parent_id = None
             flags = 0
             baggage = None
@@ -170,7 +173,7 @@ class Tracer(opentracing.Tracer):
                 span_id = parent.span_id
                 parent_id = parent.parent_id
             else:
-                span_id = self.random_id()
+                span_id = self._random_id(constants._max_id_bits)
                 parent_id = parent.span_id
             flags = parent.flags
             baggage = dict(parent.baggage)  # TODO do we need to clone?
@@ -238,7 +241,13 @@ class Tracer(opentracing.Tracer):
         self.metrics.spans_finished(1)
 
     def random_id(self):
+        """
+        DEPRECATED: use _random_id() instead
+        """
         return self.random.getrandbits(constants.MAX_ID_BITS)
+
+    def _random_id(self, bitsize):
+        return self.random.getrandbits(bitsize)
 
     def is_debug_allowed(self, *args, **kwargs):
         if not self.throttler:
--- a/tests/test_codecs.py
+++ b/tests/test_codecs.py
@@ -16,6 +16,7 @@ from __future__ import absolute_import
 
 import unittest
 from collections import namedtuple
+from itertools import product
 import six
 
 import mock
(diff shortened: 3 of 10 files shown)


bhavin192:
https://github.com/jaegertracing/jaeger-client-python/blob/73ea288218a58843f379f273a4302698600f339e/jaeger_client/sampler.py#L103-L128

Right now the `ProbabilisticSampler` depends on the fact that the `trace_id`s are 64bit. Any suggestions on how this can be implemented for 128bit trace id while maintaining compatibility for both lengths?

jaeger_client/tracer.py
```diff
@@ -237,8 +239,8 @@ def report_span(self, span):
         self.reporter.report_span(span)
         self.metrics.spans_finished(1)
 
-    def random_id(self):
-        return self.random.getrandbits(constants.MAX_ID_BITS)
+    def random_id(self, max_id_bits=constants.MAX_ID_BITS):
```
yurishkuro: This constant is no longer adequate/ misleading. Not sure why it was public in the first place either. I think we should label it deprecated and define separate constants for max bits for trace and span ids, preferably private. 

The method should not have a default.

yurishkuro:
Thanks for working on it, exciting to see this one moving ahead. Soon all clients will be 128bit capable.

jaeger_client/tracer.py
bhavin192: Something like `_max_trace_id_bits` and `_max_span_id_bits`/`_max_id_bits` in `constants.py`?

yurishkuro: yes.

bhavin192:
@yurishkuro while I'm working on the changes you suggested, the tests for Python 2.7 are failing, I think it's because tchannel uses struck.pack which does not support numbers larger than 8 bytes i.e. 64 bits

jaeger_client/thrift.py
```diff
@@ -26,6 +26,26 @@
...
+
+
+def id_to_high(big_id):
+    """
+    :param big_id: id in integer
+    :return: Returns the left most 64 bits of big_id
+    """
+    if big_id is None:
+        return None
+    return big_id >> 64
```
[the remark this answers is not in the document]
bhavin192: Thanks

bhavin192:
> For the sampler I suggest using AND to trim the ID to desired bit length.

Okay, let me try to implement that.

yurishkuro:
> if we trim down the Trace ID to 64 bits and then apply the logic of ProbabilisticSampler, I think it will not be the correct sampling rate

Sampling rate has no logical dependency on the trace IDs. However, because a trace ID is a random number, and the probabilistic sampler needs a random number to compare against the probability value, we made a choice to reuse the trace ID for that purpose. It's purely an optimization, not a logical dependency. And 64 bits as a random number it more than enough to compare against the probability value, so it's ok to use the lower-part of 128bit trace ID.

bhavin192:
@yurishkuro Aha! now I got it, seems like I misinterpreted the doc string of the function, thank you for explaining :)

yurishkuro:
feel free to edit the function doc, it's always better to have an external perspective

bhavin192:
@yurishkuro I'm done with the changes, can you please review?

jaeger_client/thrift.py
```diff
@@ -26,6 +26,24 @@
     long = int
 
 
+def id_to_low(big_id):
+    """
+    :param big_id: id in integer
+    :return: Returns the right most 64 bits of big_id
+    """
+    if big_id is not None:
+        return big_id & (_max_unsigned_id - 1)
```
yurishkuro: does Python assume None return value if there's no explicit return at the end?

jaeger_client/tracer.py
```diff
@@ -237,8 +240,8 @@ def report_span(self, span):
         self.reporter.report_span(span)
         self.metrics.spans_finished(1)
 
-    def random_id(self):
-        return self.random.getrandbits(constants.MAX_ID_BITS)
+    def random_id(self, max_id_bits):
```
yurishkuro: * for backwards compat, please don't change this function. Instead, mark it as deprecated, and use a new private one, `_random_id(bitsize)`

setup.py
```diff
@@ -60,7 +60,7 @@
             'coverage<4.4',  # can remove after https://bitbucket.org/ned/coveragepy/issues/581/44b1-44-breaking-in-ci
             'pytest-timeout==1.3.1',
             'pytest-tornado',
-            'pytest-benchmark[histogram]>=3.0.0rc1',
+            'pytest-benchmark[histogram]>=3.0.0rc1,<3.2.0',  # >=3.2.0 needs pytest>=3.8.0
```
yurishkuro: please merge master & resolve

jaeger_client/thrift.py
bhavin192: Yes. Functions in Python will return `None` if we don't return anything explicitly

bhavin192:
@yurishkuro I saw the option `gen128bit` in jaeger-client-go, that's why I used `generate_128bit_trace_id`.

jaeger_client/tracer.py
bhavin192: That makes sense, thank you.

jaeger_client/thrift.py
```diff
@@ -26,6 +26,24 @@
     long = int
 
 
+def id_to_low(big_id):
```
yurishkuro: please make these private

jaeger_client/thrift.py
```diff
@@ -26,6 +26,24 @@
...
+    """
+    :param big_id: id in integer
+    :return: Returns the right most 64 bits of big_id
+    """
+    if big_id is not None:
+        return big_id & (_max_unsigned_id - 1)
+
+
+def id_to_high(big_id):
```
yurishkuro: `_ id_to_high`

bhavin192:
@yurishkuro as far as I know the `start_span()` from tracer.py already does this i.e. sets the trace_id it gets from the context, no matter if it's a 64bit or 128bit. Do you think just updating the `test_round_trip()` from `test_codecs.py` is enough to test this compatibility?

yurishkuro:
> i.e. sets the trace_id it gets from the context, no matter if it's a 64bit or 128bit.

Well, because it never expected to see 128bit. I suggest adding the tests for the scenarios I described, and then I assume some of them will break without capturing some additional state in the span context to indicate the length of the trace id.

bhavin192:
@yurishkuro I have updated the tests to cover the scenarios, can you please take a look?

tests/test_codecs.py
```diff
@@ -453,6 +456,51 @@ def test_round_trip(tracer, fmt, carrier):
...
+        service_name='test',
+        reporter=InMemoryReporter(),
+        sampler=ConstSampler(True),
+        generate_128bit_trace_id=True)
+
+    for tracer, length in [(tracer, 16), (tracer_128bit, 32)]:
+        trace_id = (1 << 64) - 1 if length == 16 else (1 << 128) - 1
+        ctx = SpanContext(trace_id=trace_id, span_id=127, parent_id=None,
+                          flags=1)
+        span = Span(ctx, operation_name='test-%s' % fmt, tracer=None, start_time=1)
```
yurishkuro: Could you use extract from a string instead of creating the span context manually? The idea was to test passing trace id unchanged end-to-end

bhavin192: Umm, the end to end trace_id is being already tested by this modified round_trip here https://github.com/jaegertracing/jaeger-client-python/pull/230/files#diff-252cc4202700183e8f0c782fb28dbfc2R451

yurishkuro: It's not testing the hybrid mode, but perhaps we can do it later in another PR if the need arrises. AFAIK other Jaeger clients that support 128bit ID also only support a full switch mode.

yurishkuro:
@bhavin192 we recently enabled linting of tests, do you mind applying some formatting changes to make it pass?  E.g. `make lint`.

bhavin192:
@yurishkuro done :)

PR #231, opened 2019-01-10
Unpin flake8 to allow v3

yurishkuro:
and fix a few minor breaks

also remove `tests` from lint, because they were skipped in flake8 config anyway

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ clean:
 
 .PHONY: lint
 lint:
-	$(flake8) $(projects) tests
+	$(flake8) $(projects)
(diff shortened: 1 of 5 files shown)


black-adder:
> also remove `tests` from lint, because they were skipped in flake8 config anyway

I actually like linting tests... the whole codebase should conform

yurishkuro:
@black-adder I agree with you, however at the moment `tests/` does not pass flake8 lint, and it is explicitly excluded from flake8 configuration in `setup.cfg`, so it was misleading to include the dir in the lint command.

Issue #189, opened 2018-06-19
Module not found in Windows 10


fabianobr:
I need to solve a non functional requirement, to work on windows 10.

I can't work in an windows 10 environment.

Traceback (most recent call last):
  File "D:\utils\jaeger\tracer.py", line 3, in <module>
    from jaeger_client import Config
  File "C:\Python36\lib\site-packages\jaeger_client\__init__.py", line 26, in <module>
    from .tracer import Tracer  # noqa
  File "C:\Python36\lib\site-packages\jaeger_client\tracer.py", line 33, in <module>
    from .utils import local_ip
  File "C:\Python36\lib\site-packages\jaeger_client\utils.py", line 17, in <module>
    import fcntl
ModuleNotFoundError: No module named 'fcntl'

Do we have any alternative to mitigate this?

yurishkuro:
this happens when the client tries to determine IP address of the host to record it in the span. You can pass the IP in a tag to the tracer:
```
jaeger_client/constants.py:51:JAEGER_IP_TAG_KEY = 'ip'
```
then the client will not be doing the lookup. It might still blow up since utils.py imports these at the top, it could be addressed by moving imports inside `interface_ip()` function. To do this correctly it may need windows-specific code (unless there's a polyfill somewhere)

ZmeiGorynych:
Thanks! Ran into the same thing.
It seems a bit harsh for the code to fail on Windows because of this, would it not make more sense to return an IP of 'Unknown IP' or something like that? 
Happy to submit a pull request for that.

yurishkuro:
PRs are always welcome

PR #233, opened 2019-01-11
Enable windows support by isolating ioctl calls

Etienne-Carriere:
Resolves : https://github.com/jaegertracing/jaeger-client-python/issues/189

Signed-off-by: Etienne Carriere <etienne.a.carriere@socgen.com>

### DIFF
--- a/jaeger_client/utils.py
+++ b/jaeger_client/utils.py
@@ -12,12 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from six.moves import range
-
-import fcntl
 import socket
 import struct
+
 import time
+from six.moves import range
 
 
 class ErrorReporter(object):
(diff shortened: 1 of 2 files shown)


jaeger_client/utils.py
```diff
@@ -65,27 +64,56 @@ def local_ip():
...
-            i + bytes(n) for i in (b'eth', b'en', b'wlan') for n in range(3)
-        ]  # :(
-        for interface in interfaces:
-            try:
-                ip = interface_ip(interface)
+        ip = get_local_ip_by_interfaces()
+        if ip is None:
+            ip = get_local_ip_by_socket()
+            if ip is None:
+                ip = 'Unknown IP'
```
yurishkuro: maybe it's better to skip the IP tag in this case? Returning invalid string as IP might break something downstream.

Etienne-Carriere:
@yurishkuro , could you please re-review ?

PR #247, opened 2019-03-31
Support Tornado 5

condorcet:
There are minor changes related to migration from `gen.engine` (which was deprecated) to `gen.coroutine`.
Also `test_composite_reporter` has been moved to appropriate `AsyncTestCase` class due it asynchronous behavior.
As I see there is no problem to remove upper bound of Tornado, but it will lead to conflicts because some tests dependencies expected Tornado < 6 (you can see build failed on another branch https://travis-ci.org/condorcet/jaeger-client-python/jobs/513798351).
At this moment we also have inconsistent test dependencies, but it is not affected on tests:
- `tchannel` wants Tornado < 5
- `opentracing-instrumentation` (as the direct dependency and transitive dependency from `thchannel`) wants Tornado < 5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,10 @@ dist: trusty
 
 language: python
 
-matrix:
-  include:
-  - python: '2.7'
-    env: COVER=1
-  - python: '2.7'
-    env: CROSSDOCK=1
-  - python: '3.5'
-  - python: '3.6'
+python:
+  - "2.7"
+  - "3.5"
+  - "3.6"
 
 services:
   - docker
@@ -24,9 +20,21 @@ env:
     - secure: "V2CBqCBxMOv9H3v/DcDVPABcxktssVaupbQh7qU2rLFWv5SmJKBN+0ZixmPFQ5ivwNcs93HM+Qck6plY+VGxIFMppvkUveJqKoXnLvk/fdrOyeneQubYXxR3Z2FF5Klml8Jdhi6L6BbYNyeCSMBUljCHj5cnEVA9Z+z8jUxywxk1KjtEDIulzb79TaZaA6pQVA9NRzxnGWqNnlV+9D35Ik0JXMRVMumSPwux6yJvoCQG3J6oJ/drvsdD890xnofuDc71yZTNXKQABqkRheTvDFaQdP8TDSyy71q9A74L+PRpwxP+cF77w8EzXRcFGKrl0xJZMQDpr8Wz8JUL36ft9xSoI6EJChyipIl2G2K1wpx6N8/9YxTkVIDiSxt5CvqJj2gt37GjR8W6qU3cBnwrpFaC+RJtiou/YJstSmwC/hsskTSKy3wDaa7G13pMhXEAhvfplDOchOuNG6ElLkatJ1RxPTdKtZc1YGg4OTcXFg+BqRIEZ57RJ3qTsxpWjZYmo2RwklAL6CfaFBUPGZiNrE4MGhO4/cXbqqaXG+X2sercW658+o7M7RE/BwC4+e5mDGOLGiry5Qy7+9bVJqJKTN8UVmhCJ23MNasVqAcQvzsGp1nZYQYFzcSfGRzqrqM2zIY6sjj5rfDGyYd8ifQFEi6fdWgMqF17PgUvKUHd5dM="
     # DOCKER_PASS
     - secure: "xxoMQo7hF+yr8SyITXonPQahEDcy5Y+uyX4kiVQxtlRIN/Nd1vL64PIocB/M4k9MhSMmz15fFjTtHuzk/tVKpHpgX9SFVCRRAt5qbzekm9NOce25BqJm33zgEKO979sNtn5Aen8ocU5uIar9IKhzV/SaEXPDqNmCGlnF/xAbvKjtTq+PjLy4xj6UKHTvSfYQFsOBgVzmZk759XVAiODN2szIzzXTNMfxOK6FG7w7O2U9VZ7lsvqC7NymXSocp4iF4kOlvAq13sFGYYMJ+1yReODAGNsUZv0TPgPT5NSxPA5OxjVNlPnlmvD6IjZvzpkciPyjsUTV8OIiuyE8GdUKc1FHKvUavN16ac7vC2dAv1VYNFvlr61RIpUAMNk+7k9fX6M5mHBzhnfNSjF2MmKyVDv4PqgLAQOtk9Vbhd8EdZS+/3MRg8DZDWQE8h0rPJQTPWDTMQ4q4yQcpygBnICB0gACIlTFSevnvsTQGUQfR/0GPB8AGsEHTSw7FUQI7dD1JYteRlgk9vNeYpeYrtgom5v9KSeaio1gTB7OlTd3car2iTYSPUgYpXLLY4vV/HEVW+FUWtCkgojWACUmFschNGPxV+nkjjlgj+Nonnz4XK5Y4l4W8Y6N7fTpxEIB9O+5e+Xnf6RhvoWvw/uMbagvK2U+sqFcV0tj8KyfWqtAm3s="
+  matrix:
+    - TORNADO=">=4,<5"
+    - TORNADO=">=5,<6"
+    - COVER=1 CROSSDOCK=1
+
+matrix:
+  exclude:
+  - python: "3.5"
+    env: COVER=1 CROSSDOCK=1
+  - python: "3.6"
+    env: COVER=1 CROSSDOCK=1
 
 install:
   - make bootstrap
+  - pip install -q "tornado$TORNADO"
   - if [ "$CROSSDOCK" == "1" ]; then bash ./scripts/install-crossdock-deps.sh ; fi
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,6 @@ dist: trusty
 
 language: python
 
-python:
-  - "2.7"
-  - "3.5"
-  - "3.6"
-
 services:
   - docker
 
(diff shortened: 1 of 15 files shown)


condorcet:
It seems that `opentracing_instrumentation` isn't necessary, so I removed it from test dependencies.

yurishkuro:
It is used in the integration tests:  https://github.com/jaegertracing/jaeger-client-python/blob/cb507453948a780cdfb1bc845090ec502b5e52d1/crossdock/server/server.py#L27

setup.py
```diff
@@ -39,7 +39,7 @@
     install_requires=[
         'threadloop>=1,<2',
         'thrift',
-        'tornado>=4.3,<5',
+        'tornado>=4.3,<6',
```
yurishkuro: I am not comfortable with this change as is considering that it will force CI to install v5, meaning we'd stop testing with v4. Since Tornado is (unfortunately) a critical dependency, we need to extend CI to test with both versions, either by using tox or by manually installing 4.x before `make bootstrap`

tests/test_reporter.py
```diff
@@ -296,3 +260,40 @@ def send(_):
         yield reporter.close()
         assert reporter.queue.qsize() == 0, 'all spans drained'
         assert count[0] == 4, 'last span submitted in one extrac batch'
+
+    @gen_test
+    def test_composite_reporter(self):
```
yurishkuro: why does this function need to move?

condorcet:
It seems that inconsistent dependencies I mentioned above broke the integration tests.
```
ERROR: opentracing-instrumentation 3.0.1 has requirement tornado<5,>=4.1, but you'll have tornado 5.1.1 which is incompatible.
ERROR: tchannel 1.3.1 has requirement opentracing<2,>=1.1, but you'll have opentracing 2.2.0 which is incompatible.
ERROR: tchannel 1.3.1 has requirement opentracing-instrumentation<3,>=2.3, but you'll have opentracing-instrumentation 3.0.1 which is incompatible.
ERROR: tchannel 1.3.1 has requirement tornado<5,>=4.3, but you'll have tornado 5.1.1 which is incompatible.
```
So we have to adapting `tchannel` to new versions of these libraries.

Last time when I removed `opentracing-instrumentation` from tests dependencies, pip ultimately installed inconsistent dependencies that way that they worked well https://travis-ci.org/jaegertracing/jaeger-client-python/jobs/531718682#L1080 (see `opentracing-2.2.0`, `opentracing-instrumentation-2.4.3`). That's why previous build passed :)

setup.py
condorcet: I agree, it's a good point.

nicholasamorim:
I have been testing this and so far it works fine. The actual changes are pretty small. How can we move this forward?

yurishkuro:
@nicholasamorim see https://github.com/jaegertracing/jaeger-client-python/pull/247#discussion_r283918803

yurishkuro:
@condorcet there's very little chance tchannel will be upgraded, that project is in maintenance mode. Considering that it is only used in crossdock integration tests, it doesn't need to be upgraded as long as those tests are still run with tornado 4

condorcet:
@yurishkuro I thought about this way but I don't understand is it possible now: in Dockerfile we make environment with dependencies from the `requirements ` files of the project https://github.com/jaegertracing/jaeger-client-python/blob/master/crossdock/Dockerfile#L18 and install `jaeger-client` from sources. So before integration tests we already have `jaeger-client` with latests dependencies and `tchannel` doesn't compatible with them. And it including not only Tornado, but newest "opentracing" libraries too.
Actually I've started explore tchannel library and I hope it doesn't take so much time :)

tests/test_reporter.py
condorcet: They made some changes in `Tornado 5+` (for asyncio compatibility) and done callback of future will fire on next loop cycle. That's why I made `yield f` to resolving future and it made the test asynchronous.

condorcet:
@yurishkuro I've explored `tchannel` library and unfortunately it's not so easy to support Tornado 5 with the latest opentracing libraries (one of the reason is that the library tests depends on `jaeger-client` itself :) But good news that I made fix of integration test, despite inconsistent test dependencies.

setup.py
condorcet: Done. Using `tox` is better way to solve this problem, but I think that environment matrix in .travis.yml enough for the PR.

jaeger_client/throttler.py
```diff
@@ -115,8 +115,7 @@ def callback():
         periodic = PeriodicCallback(
             callback=callback,
             # convert interval to milliseconds
-            callback_time=self.refresh_interval * 1000,
-            io_loop=self.channel.io_loop)
```
yurishkuro: please explain why `io_loop=self.channel.io_loop` is removed

setup.py
```diff
@@ -67,7 +67,7 @@
             'flake8-quotes',
             'codecov',
             'tchannel>=0.27', # This is only used in python 2
-            'opentracing_instrumentation>=2,<3',
+            'opentracing_instrumentation',
```
yurishkuro: I prefer to keep the ranges, e.g. previously`>=1,<2` was not compatible with new opentracing-2.x

.travis.yml
```diff
@@ -24,9 +20,26 @@ env:
...
     # DOCKER_PASS
     - secure: "xxoMQo7hF+yr8SyITXonPQahEDcy5Y+uyX4kiVQxtlRIN/Nd1vL64PIocB/M4k9MhSMmz15fFjTtHuzk/tVKpHpgX9SFVCRRAt5qbzekm9NOce25BqJm33zgEKO979sNtn5Aen8ocU5uIar9IKhzV/SaEXPDqNmCGlnF/xAbvKjtTq+PjLy4xj6UKHTvSfYQFsOBgVzmZk759XVAiODN2szIzzXTNMfxOK6FG7w7O2U9VZ7lsvqC7NymXSocp4iF4kOlvAq13sFGYYMJ+1yReODAGNsUZv0TPgPT5NSxPA5OxjVNlPnlmvD6IjZvzpkciPyjsUTV8OIiuyE8GdUKc1FHKvUavN16ac7vC2dAv1VYNFvlr61RIpUAMNk+7k9fX6M5mHBzhnfNSjF2MmKyVDv4PqgLAQOtk9Vbhd8EdZS+/3MRg8DZDWQE8h0rPJQTPWDTMQ4q4yQcpygBnICB0gACIlTFSevnvsTQGUQfR/0GPB8AGsEHTSw7FUQI7dD1JYteRlgk9vNeYpeYrtgom5v9KSeaio1gTB7OlTd3car2iTYSPUgYpXLLY4vV/HEVW+FUWtCkgojWACUmFschNGPxV+nkjjlgj+Nonnz4XK5Y4l4W8Y6N7fTpxEIB9O+5e+Xnf6RhvoWvw/uMbagvK2U+sqFcV0tj8KyfWqtAm3s="
+  matrix:
+    - TORNADO=">=4,<5"
+    - TORNADO=">=5,<6"
+    - COVER=1
+    - CROSSDOCK=1
```
yurishkuro: would this cause tornado-5 w/ crossdock tests? I would rather keep them at v4. I am not sure our unit tests are actually exercising many of the thrift-generated files, and that's one of the compatibility questions that worry me:
```
-  @gen.engine
+  @gen.coroutine
```

jaeger_client/throttler.py
condorcet: `io_loop` argument was deprecated and removed from Tornado 5+ (http://www.tornadoweb.org/en/stable/releases/v5.0.0.html#tornado-ioloop):

> The io_loop argument to PeriodicCallback has been removed.
> It is now possible to create a PeriodicCallback in one thread and start it in another without passing an explicit event loop.

But it's not obvious here, because `io_loop` that we are using here was made and started in another thread via `threadloop` (https://github.com/jaegertracing/jaeger-client-python/blob/master/jaeger_client/local_agent_net.py#L71) and we should pass the loop explicitly.
Fortunately we are starting polling through callback of the `io_loop` (https://github.com/jaegertracing/jaeger-client-python/blob/master/jaeger_client/throttler.py#L103 and https://github.com/jaegertracing/jaeger-client-python/blob/master/jaeger_client/sampler.py#L399) and the callback will be invoked in the thread of the `io_loop`. And that's why `PeriodicCallback` will take right loop to start (it takes existing loop from current thread), instead of creating new right in main thread.

yurishkuro: > argument was deprecated and removed from Tornado 5+ 

what about people still using it w/ Tornado 4?

condorcet: This argument is optional in Tornado 4: if no `io_loop` then use current loop https://github.com/tornadoweb/tornado/blob/branch4.5/tornado/ioloop.py#L998 Current loop will be taken from current thread and in our case it will be thread with already instantiated threadloop.

.travis.yml
condorcet: Done.

condorcet:
@yurishkuro I found that installing correct version of Tornado from travis environment doesn't work properly, because `test-import` (https://github.com/jaegertracing/jaeger-client-python/blob/master/Makefile#L41) reinstall jaeger-client dependencies in new virtual environment. So I've added installing Tornado right into this target.
Now you can see that correct version of Tornado will be installed before tests:
- python 2.7, tornado 4, crossdock https://travis-ci.org/jaegertracing/jaeger-client-python/jobs/541331259#L1343
- python 2.7, tornado 4 https://travis-ci.org/jaegertracing/jaeger-client-python/jobs/541331260#L845
- python 2.7, tornado 5 https://travis-ci.org/jaegertracing/jaeger-client-python/jobs/541331261#L829
- python 3.5, tornado 4 https://travis-ci.org/jaegertracing/jaeger-client-python/jobs/541331262#L762
- python 3.5, tornado 5 https://travis-ci.org/jaegertracing/jaeger-client-python/jobs/541331263#L749
- python 3.6, tornado 4 https://travis-ci.org/jaegertracing/jaeger-client-python/jobs/541331264#L784
- python 3.6, tornado 5 https://travis-ci.org/jaegertracing/jaeger-client-python/jobs/541331265#L771

setup.py
condorcet: Done.

setup.py
yurishkuro: adding `tox` support would be great, if you feel like it (doesn't have to be this PR)

yurishkuro:
scary changes, but ok. This might come back to bite us:
```
Successfully built tornado
ERROR: tchannel 1.3.1 has requirement opentracing<2,>=1.1, but you'll have opentracing 2.2.0 which is incompatible.
ERROR: tchannel 1.3.1 has requirement opentracing-instrumentation<3,>=2.3, but you'll have opentracing-instrumentation 3.0.1 which is incompatible.
Installing collected packages: tornado
  Found existing installation: tornado 5.1.1
    Uninstalling tornado-5.1.1:
      Successfully uninstalled tornado-5.1.1
Successfully installed tornado-4.5.3
```

pavank:
I have other libraries using tornado >5 and uninstalling latest tornado versions and installing a lower version breaks things on existing projects.

Issue #163, opened 2018-04-23
Stacktrace is not set properly on errors


nziebart:
The default opentracing impl logs the stacktrace in the event of an exception:
https://github.com/opentracing/opentracing-python/blob/master/opentracing/span.py#L199

However the stacktrace object gets converted to a string via `str(tb)` and it results in the following display in the UI: `python.exception.tb | "<traceback object at 0x7f7a2aad9638>"`

I think it might actually be easiest to fix this in opentracing, by logging a string representation of the stacktrace rather than the object

yurishkuro:
I think the right place to fix it is in the reporter/converter. Pushing the conversion upstream to opentracing means the work will have to be done every time, even though the trace may be unsampled.

SEJeff:
`traceback.format_exc()` is the magic to get the last raised exception as a string. The code in opentracing-python (that returns `repr(exception_object)` seems useless otherwise, but I'd expect the value to be truncated past a certain number of chars.

nziebart:
Yeah, I think the main question here is where should we call `traceback.format_exc()`. Currently there is no real concept of a value converter, it just goes to [this function](https://github.com/jaegertracing/jaeger-client-python/blob/90ddf76da60b8a3304ea30a5ff08e188eb2f2de6/jaeger_client/thrift.py#L39) which does `str(val)`

The easy fix would be to make this conversion logic know about traceback objects and special case them. Alternatively we could let the caller pass in a callable/lambda that returns the string, which we call only if the trace is sampled.

PR #250, opened 2019-06-06
Add stack traceback to error logs

qqi0O0:
Previously, error logs only come with a tag of the form
"traceback object at location". Now they also report the actual
stack trace.

Resolves: #163

Signed-off-by: Qi Qi <piano.keys0@gmail.com>


qqi0O0:
Two tests have been added. I also modified the behavior to be more logical (stack reported under 'stack' tag rather than 'error.object' tag).

qqi0O0:
Actually, does it make any sense to include the message about the traceback location? Should I just use the traceback message and get rid of the first line of "traceback object at location"?

jaeger_client/thrift.py
```diff
@@ -90,6 +97,18 @@ def make_tag(key, value, max_length):
         )
 
 
+def _make_traceback_tag(key, value, max_length):
+    key = _to_string(key)
+    value = traceback.format_exc()
```
yurishkuro: This looks weird - you're ignoring passed `value` and using some global state?

tests/test_traceback_log.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_traceback_full():
+    """Test that a traceback is logged with both location and message"""
+    mock_tracer = mock.MagicMock()
+    mock_tracer.max_tag_value_length = 300
+    context = SpanContext(trace_id=1, span_id=2, parent_id=None, flags=1)
+    span = Span(context=context, operation_name='traceback_test', tracer=mock_tracer)
+
+    try:
+        raise ValueError('Something unexpected happened!')
```
yurishkuro: what about using context manager properties of the span to catch the exception automatically?
```
with span:
    raise X
```
The base class implements it: https://github.com/opentracing/opentracing-python/blob/master/opentracing/span.py#L218

jaeger_client/thrift.py
qqi0O0: Good point, I will format the traceback object `value` instead.

tests/test_traceback_log.py
qqi0O0: That's a good point, I've made that change to tests.

jaeger_client/thrift.py
```diff
@@ -82,6 +83,12 @@ def make_tag(key, value, max_length):
             key=key,
             value=value
         )
+    elif type(value).__name__ == 'traceback':
```
yurishkuro: is every possible value guaranteed to have `type(value).__name__` expression working?

jaeger_client/thrift.py
```diff
@@ -90,6 +97,18 @@ def make_tag(key, value, max_length):
         )
 
 
+def _make_traceback_tag(key, value, max_length):
+    key = _to_string(key)
+    value = ''.join(traceback.format_tb(value))
+    if len(value) > max_length:
+        value = value[:max_length]
```
yurishkuro: Hmm, max_length may be pretty limiting for a stack trace. Would it make sense to have a different parameter controlling max stack trace size?

jaeger_client/thrift.py
qqi0O0: I'm not completely sure it's guaranteed, but line 71 uses the same expression to check for booleans, so the newly added line shouldn't cause any new issues.

yurishkuro: +1

jaeger_client/thrift.py
qqi0O0: I did notice stack traces getting cut off rather early, so it would definitely make sense to extend that length for stack traces. It seems like a lot would need to be changed in order to add the max stack trace size as a parameter though. I can make those changes if adding such a parameter feels like a good idea. An alternative would be to use a larger constant instead of having it as a parameter.

jaeger_client/thrift.py
```diff
@@ -90,6 +97,18 @@ def make_tag(key, value, max_length):
         )
 
 
+def _make_traceback_tag(key, value, max_length):
```
yurishkuro: I would prefer `max_length=constants. MAX_TRACEBACK_LENGTH` and then remove the new parameter from all "process" related methods, because process tags should never have a stack trace in them, and it makes that API look weird all the way up.

qqi0O0: Done. I'm not sure how to test the behavior of traceback tags getting cut off now though.

yurishkuro:
I  was referring to "process"-related functions only. This one can still take the 3rd parameter as you had before:
```python
def make_tag(key, value, max_length):
```

qqi0O0:
> I was referring to "process"-related functions only. This one can still take the 3rd parameter as you had before:
> 
> ```python
> def make_tag(key, value, max_length):
> ```

I'm not totally sure how far up we want to have the parameter `max_traceback_length`. Should it be a parameter in `make_tag`? `make_tags`? `make_log`?

yurishkuro:
The correct way is to mirror `MAX_TAG_VALUE_LENGTH`, which starts from Config object, 

https://github.com/jaegertracing/jaeger-client-python/blob/66437fa14a69ce08a5ed4f3e7af1d6c94f7753e8/jaeger_client/config.py#L46

then to Tracer 

https://github.com/jaegertracing/jaeger-client-python/blob/5a237ce27d37b4a0a1a6bde531d85003d677e547/jaeger_client/tracer.py#L53

then used by Span

https://github.com/jaegertracing/jaeger-client-python/blob/9771cc9a486e1b72ccece7efd44ae5e1befb755f/jaeger_client/span.py#L85-L88

So yes, I would expect it to be passed to `thrift.make_tag()`. However, I would not pass it to `reporter.set_processs()`, because process tags are not expected to contain stack traces.

Issue #251, opened 2019-06-22
Bug: span.finish() reports duplicate spans


jan25:
Current `span.finish()` [implementation](https://github.com/jaegertracing/jaeger-client-python/blob/master/jaeger_client/span.py#L60) just keeps tracks of `end_time`, and reports span on each call which will result in duplicate spans under same trace.

Expected behavior would be- only report the span the first `finish()` is called and fail from second time onwards. Maybe we just need a simple check if the `end_time` was in the past?

This can be reproduced with const sampler, other kind of sampler might skip reporting duplicate spans

yurishkuro:
OpenTracing leaves tracer behavior as undefined if span.finish() is called >1 times, so technically this is not a bug. Yes, we may want to put a guard and log an error message if that happens. But you should really look into your instrumentation and fix it to not call finish twice.

jan25:
Thanks! Yes, my instrumentation was buggy. The code didn't actually call `span.finish()` twice, but one of calls was indirect because i did `scope.close()`. I might've missed this spec for [scope_manager](https://github.com/opentracing/opentracing-python/blob/master/opentracing/scope_manager.py#L42).

If having a guard in `span.finish()` is not very useful we can close this issue.

yurishkuro:
I think having a guard is useful, we have it in some other Jaeger clients.

condorcet:
Hm, I found that `span.finish` behavior is inconsistent now. If span isn't sampled it won't be finished in term of flags `finished` and `end_time` https://github.com/jaegertracing/jaeger-client-python/blob/master/jaeger_client/span.py#L74 I don't think it's a big problem but it may be uncomfortable  when debugging or for third-party tracing library, for example. cc @jan25

jan25:
Hi @condorcet ! Semantically this behavior of not finishing isn't changed from before this issue/PR. And, span isn't sampled means the respective trace(all spans under it) won't be reported, so marking span as finished or trying to report would be redundant anyway.

Could you explain more on what you're trying to achieve/debug w.r.t non sampled spans?

PR #253, opened 2019-06-28
Keep track if span is finished

jan25:
This PR adds a guard to `span.finish()` function to check if span has finished, and not report a span more than once.
We do have similar check in [java client](https://sourcegraph.com/github.com/jaegertracing/jaeger-client-java/-/blob/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java#L179) for example.

Fixes https://github.com/jaegertracing/jaeger-client-python/issues/251

Signed-off-by: Abhilash Gnan <abhilashgnan@gmail.com>

### DIFF
--- a/jaeger_client/span.py
+++ b/jaeger_client/span.py
@@ -24,8 +24,8 @@ from opentracing.ext import tags as ext_tags
 from . import codecs, thrift
 from .constants import SAMPLED_FLAG, DEBUG_FLAG
 
-
 logger = logging.getLogger('jaeger_tracing')
+
 
 class Span(opentracing.Span):
     """Implements opentracing.Span."""
--- a/jaeger_client/span.py
+++ b/jaeger_client/span.py
@@ -24,8 +24,8 @@ from opentracing.ext import tags as ext_tags
 from . import codecs, thrift
 from .constants import SAMPLED_FLAG, DEBUG_FLAG
 
-logger = logging.getLogger('jaeger_tracing')
 
+logger = logging.getLogger('jaeger_tracing')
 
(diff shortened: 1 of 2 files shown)


jaeger_client/span.py
```diff
@@ -67,11 +71,15 @@ def finish(self, finish_time=None):
         :param finish_time: an explicit Span finish timestamp as a unix
             timestamp per time.time()
         """
-        if not self.is_sampled():
-            return
+        with self.update_lock:
```
yurishkuro: I suggest keeping the is_sampled check above. Otherwise you're introducing a mutex lock for all spans, even unsampled ones, which is an unnecessary performance hit.

jan25: Good point, moved it above. So i suppose we're ok with not marking span as finished if its not sampled firstly.

jan25:
Thanks for the review!

Added tests for `span.finish()`. I couldn't add find a nice way to test for warning in log with our current tests setup. I did manual testing though, and the warning did show in the logs.

jaeger_client/span.py
```diff
@@ -37,6 +40,7 @@ def __init__(self, context, tracer, operation_name,
         self.operation_name = operation_name
         self.start_time = start_time or time.time()
         self.end_time = None
+        self.finished = False
```
yurishkuro: please add `finished` to `__slots__` above

jan25: Oh, didn't notice that. Done!

yurishkuro:
Thanks you!

PR #262, opened 2019-10-09
Remove Tornado upper boundary

condorcet:
As I can see we don't have any problem with Tornado 6 because it use asyncio under the hood like Tornado 5 in python 3.
About tests:
At this moment we use `TornadoScopeManager` in `test_crossdock.py` for tests that only run in python 2 (because of `tchannel`). Tornado 6 doesn't have `stack_context` anymore so we get import error.
For other few tests in this file scope manager and tracer unnecessary, so we can make conditional import of `TornadoScopeManager` only for python 2.
Maybe when this tests be ready for python 3, it's better to use `AsyncioScopeManager` for them.

### DIFF
--- a/tests/test_crossdock.py
+++ b/tests/test_crossdock.py
@@ -20,7 +20,6 @@ import json
 import os
 import pytest
 import opentracing
-from opentracing.scope_managers.tornado import TornadoScopeManager
 from mock import MagicMock
 from tornado.httpclient import HTTPRequest
 from jaeger_client import Tracer, ConstSampler
@@ -29,6 +28,10 @@ from crossdock.server.endtoend import EndToEndHandler, _determine_host_port, _pa
 
 if six.PY2:
     from crossdock.server import server
+    from opentracing.scope_managers.tornado import TornadoScopeManager
+    scope_manager = TornadoScopeManager
+else:
+    scope_manager = None
 
 tchannel_port = '9999'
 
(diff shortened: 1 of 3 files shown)


tests/test_crossdock.py
```diff
@@ -48,7 +51,7 @@ def tracer():
         service_name='test-tracer',
         sampler=ConstSampler(True),
         reporter=InMemoryReporter(),
-        scope_manager=TornadoScopeManager()
+        scope_manager=scope_manager() if scope_manager else None
```
Jamim:
```suggestion
        scope_manager=scope_manager and scope_manager()
```

yurishkuro: why not just instantiate it next to the import?

yurishkuro:
TornadoScopeManager comes from opentracing. Are you saying it's not compatible w/ Tornado-6? Maybe a change needs to be made there, rather than conditionally importing here.

condorcet:
> TornadoScopeManager comes from opentracing. Are you saying it's not compatible w/ Tornado-6? Maybe a change needs to be made there, rather than conditionally importing here.

Hi @yurishkuro Yes, it's true, `TornadoScopeManager` that based on `stack_context` works in Tornado<6. Tornado 6 became asyncio framework and `stack_context` has been removed. We discussed about this problem in OT https://github.com/opentracing/opentracing-python/pull/118
We should take into account this, if we want to adapt tests for python 3 with Tornado 6. For example, use here `AsyncioScopeManager` or `ContextVarsScopeManager` (if approved in OT).

But now, we can just skip importing `TornadoScopeManager` in tests.

yurishkuro:
ok, I am not opposed to this change.

fenglixa:
Which version of jaeger-client involved this fixes?

yurishkuro:
4.2.0

PR #269, opened 2019-12-01
Allow adding 'service' label to Prometheus metrics

alexppg:
Add support for the prometheus metrics to have a label called `service`, which can be used to identify every microservice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ if __name__ == "__main__":
                 'param': 1,
             },
             'logging': True,
-        },  
+        },
         service_name='your-app-name',
         validate=True,
     )
@@ -67,7 +67,7 @@ if __name__ == "__main__":
                 'reporting_port': 'your-reporting-port',
             },
             'logging': True,
-        },  
+        },
         service_name='your-app-name',
         validate=True,
     )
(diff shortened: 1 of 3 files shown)


alexppg:
Hey! Thanks for the comments. Why you want to default it to none? I've default it to an empty string because is the same is done with namespace. Also, why you say it's assumed that it exists? It doesn't. @yurishkuro.

yurishkuro:
I don't know why namespace is defaulted to empty string, maybe it was a requirement of Prom client, or just an oversight. We don't need to follow that pattern. None is the idiomatic way to indicate that the parameter was not passed, especially when the receiver is planning not to use it in that case.

The reason I say it's "required" in the current PR is because the label is always added to the metric, regardless of the value passed. To be backwards compatible it should only be added if the value is not None.

alexppg:
Thanks for your explanations! I've just pushed most of the changes. I haven't pushed the one you say to make it DRYer. 
The prometheus metrics needs the labels when declaring the object (gauge or counter) and also when updating the metric.
So, since both the declaration and the update happens in the create_counter and create_gauge, I think it's the only way to do it, but if you see another way, I'll do it gladly.

yurishkuro:
Advice for the future: do not create pull requests from `master` branch of your fork, always create a feature branch first. Your master should ideally always match the upstream master.

alexppg:
Ok, thank you very much for your advices and commits!

alexppg:
ping @yurishkuro

yurishkuro:
released 4.3.0

alexppg:
Thanks!

Issue #273, opened 2020-01-29
changed default collector sampling probability does not propagate correctly


tjb900:
## Requirement - what kind of business use case are you trying to solve?

We have long-lived applications using the python client library, with a centralised strategies file served from the collector, and most operations use the default probability; only a handful are overridden with per-operation strategies.

We'd like to be able to (and assumed we could) change this default probability and have the overall trace volume respond accordingly.  However, I think there is a bug that means this isn't quite the case.

## Problem - what in Jaeger blocks you from solving the requirement?

The way `AdaptiveSampler` works appears to be that at `is_sampled` time, operations without a specified strategy create a new per-operation `GuaranteedThroughputProbabilisticSampler` with the (current) default probabilities:

https://github.com/jaegertracing/jaeger-client-python/blob/2bd1319e302c033eb3341ee50eaeadc010853e8d/jaeger_client/sampler.py#L290

Later, when `AdaptiveSampler.update` is called, the default probability is changed, and the default sampler re-created.  However, all of the per-operation `GuaranteedThroughputProbabilisticSampler`s that were created with the previous default remain in place, unchanged:

https://github.com/jaegertracing/jaeger-client-python/blob/2bd1319e302c033eb3341ee50eaeadc010853e8d/jaeger_client/sampler.py#L311

## Proposal - what do you suggest to solve the problem or improve the existing situation?

I think that the dynamically-created per-operation samplers in `AdaptiveSampler` should remember that they were spawned dynamically using the default, and should be regenerated when that default changes.  Will submit a PR with that change to accompany this issue.

PR #274, opened 2020-01-29
Keep default per-operation samplers separate

tjb900:
Test to demonstrate, and one possible fix for, #273

### DIFF
--- a/jaeger_client/sampler.py
+++ b/jaeger_client/sampler.py
@@ -269,6 +269,7 @@ class AdaptiveSampler(Sampler):
             samplers[operation] = sampler
 
         self.samplers = samplers
+        self.per_operation_default_samplers = {}
         self.default_sampler = \
             ProbabilisticSampler(strategies.get(DEFAULT_SAMPLING_PROBABILITY_STR,
                                                 DEFAULT_SAMPLING_PROBABILITY))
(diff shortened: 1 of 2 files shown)


jaeger_client/sampler.py
```diff
@@ -280,14 +281,19 @@ def __init__(self, strategies, max_operations):
     def is_sampled(self, trace_id, operation=''):
         sampler = self.samplers.get(operation)
         if not sampler:
-            if len(self.samplers) >= self.max_operations:
+            sampler = self.per_operation_default_samplers.get(operation)
+        if not sampler:
+            if (
+                len(self.samplers) + len(self.per_operation_default_samplers) >=
```
tjb900: note for reviewers: with the current configuration flake8 complains whether the line break is either before _or_ after this operator.  I think this means that either W503 or W504 needs to be disabled depending on the project's preference, or `ignore` should be changed to `extend-ignore`?   https://gitlab.com/pycqa/flake8/issues/466

tjb900:
Thanks for the review.

> Should B still apply previously configured probability, or should it switch to the default?

Hmmm yes good point.  The intent of the modified strategies file is presumably that all operations not named explicitly should use the default, so I guess they should switch back to that.  If that's your thinking too, I'll update to make that happen.

jaeger_client/sampler.py
tjb900: @yurishkuro Any thoughts on which way to go on this one?

yurishkuro: You can extract the longest expression into a variable.

tjb900: I guess I meant more in terms of fixing the mutually contradictory configuration of the style checking in CI.  As you say though, I can certainly work around it in this case.

Issue #272, opened 2020-01-20
Getting error ModuleNotFoundError: No module named 'jaeger.ttypes'


aroopghoshtarento:
in python3.7 while running i am getting below error 
```
from jaeger_client import Config
  File "/Users/aroop/python/env/lib/python3.7/site-packages/jaeger_client/__init__.py", line 27, in <module>
    from .config import Config  # noqa
  File "/Users/aroop/python/env/lib/python3.7/site-packages/jaeger_client/config.py", line 26, in <module>
    from .reporter import (
  File "/Users/aroop/python/env/lib/python3.7/site-packages/jaeger_client/reporter.py", line 32, in <module>
    from jaeger_client.thrift_gen.agent import Agent
  File "/Users/aroop/python/env/lib/python3.7/site-packages/jaeger_client/thrift_gen/agent/Agent.py", line 13, in <module>
    from .ttypes import *
  File "/Users/aroop/python/env/lib/python3.7/site-packages/jaeger_client/thrift_gen/agent/ttypes.py", line 12, in <module>
    import jaeger.ttypes
ModuleNotFoundError: No module named 'jaeger.ttypes'
```

yurishkuro:
I've seen this long time ago. Are you running some special form of Python with restrictions that block the following code from taking effect?

https://github.com/jaegertracing/jaeger-client-python/blob/2bd1319e302c033eb3341ee50eaeadc010853e8d/jaeger_client/__init__.py#L19-L23

aroopghoshtarento:
no. I am using virtualenv and then running the python application inside env with flask framework

yurishkuro:
We are not currently testing with 3.7, so it might be an issue there, or something specific to your setup.

I changed the title to be about 3.7, since we should be supporting it.

yurishkuro:
Well, ok, py3.7 tests are fine. I suspect it's something about your env. Try to present a reproducer.

aroopghoshtarento:
below are my env dependency in requirements.txt file

APScheduler==3.6.1
argcomplete==1.8.2
astroid==2.2.5
atomicwrites==1.3.0
attrs==19.1.0
autopep8==1.4.4
backcall==0.1.0
beautifulsoup4==4.5.3
bleach==3.1.0
blis==0.4.1
cachetools==3.1.1
certifi==2019.3.9
chardet==3.0.4
Click==7.0
contextlib2==0.5.5
coverage==4.5.4
cymem==2.0.2
[...]

PR #275, opened 2020-02-02
Test with Python 3.7

yurishkuro:
Issue #272


yurishkuro:
Fails to install pycurl
```
    In file included from src/docstrings.c:4:0:
    src/pycurl.h:191:30: fatal error: gnutls/gnutls.h: No such file or directory
```

Issue #279, opened 2020-05-25
Suggestion for improvement - python logging integration


lukeplausin:
## Requirement - what kind of business use case are you trying to solve?

I've recently done some work instrumenting an application to use this library for tracing. Whilst I was doing this I developed a couple of custom features which I think could be helpful for others, so I would like to contribute back to the project.

## Problem - what in Jaeger blocks you from solving the requirement?

In my use case it is important to capture logging output with the spans in order to make sense of what the application is doing. Since the project already uses the python logger extensively, and I didn't want to refactor all of the logging statements.

I like the `with tracer.start_span` mnemonic for opening spans, however once you have the span open, if you want to add a log statement inside the span you must explicitly call `span.log_kv` inside the span.

```
logger = logging.getLogger(__name__)
with tracer.start_span('TestSpan') as span:
    span.log_kv({'event': 'test message', 'life': 42})
    logger.debug('test message')
```

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Since my project already uses the python logging module, I decided to write a similar mnemonic to automatically capture output from python logging objects whilst within the span.

My log handling class uses the `with` mnemonic to add a custom log handler to the logger object whilst inside the scope of the current span. I've tested it with my application and it seems to work nicely.

```
logger = logging.getLogger(__name__)
with tracer.start_span('TestSpan') as span:
    with MyLogHandler(span=span, logger=logger) as lh:
        logger.debug('test message')
```

I think it would be really nice if it was possible to provide this functionality to devs with the jaeger client out of the box. If there was an optional `logger` argument that could do this whilst creating the span then it would make the process even more streamlined.

```
logger = logging.getLogger(__name__)
with tracer.start_span('TestSpan', logger=logger) as span:
    logger.debug('test message')
    # log will be automatically sent to the tracing client under the current span
```

## Any open questions to address

I already have the code from `MyLogHandler` ready to go. Do you think it would be a useful addition? Any comments, suggestions, feedback are welcome.

yurishkuro:
We cannot make changes to the OpenTracing API, so this approach isn't going to work.

I think there is a much simpler solution. You can add a static handler to the loggers which will access the current span from the thread-local storage using `tracer.scope_manager.active_span`. This way no changes to the tracing code are required at all.

PR #280, opened 2020-05-25
Added python logging autologger

lukeplausin:
#279 

The goal of this pull request is to give developers an easy way of capturing logging output from their instrumented applications using the Jaeger client library.

This change will add a new optional parameter when creating `Span` objects - `span_logger`. This may be set to either a python `logger` object, or `None`. If set to a python logger object, messages sent to that logger will also be sent to the tracer under the current span, whilst the span is active. This is achieved using a custom logging handler to capture emit events. Once the span is closed, the the custom log handler is removed from the logging object and log entries are no longer sent to the tracer.

- Added unit test
- Added a section in `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,44 @@ if __name__ == "__main__":
     )
 ```
 
+If your project makes use of the python logger, you can set up your span objects to automatically capture the output of python loggers whilst they are active.
+
+```python
+
+import logging
+import time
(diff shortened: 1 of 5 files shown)


lukeplausin:
I can't seem to get my first commit signed, I will close this PR and reopen.

Issue #50, opened 2017-05-25
Tracer.close() is not fully synchronous


yurishkuro:
It was meant to be fully synchronous to guarantee that all spans are flushed, however it only blocks until the spans are handed off to the sender, which itself is async, so if `close()` is used from a command line tool or a hello-world app that exist immediately, the spans may not have time to flush to UDP port.

kgrvamsi:
@yurishkuro do we need to append this `time.sleep(2)` when we write a production level application ?
Consider the scenario of distributed system where Agent/Collector/Query runs on Centralized host and we have code running on containers(roughly 30+) will the trace close needed based on the network lag for all these containers communicating to the centralized server?

yurishkuro:
If your services support graceful shutdown they yes, you might need to add this sleep after calling tracer.close(). But most long-running server applications are simply killed when they are upgraded or relocated, so in practice this rarely matters since close() won't be called anyway. You might lose some spans not flushed yet.

sspiros:
Hello,
I am hijacking this thread since I am facing an issue with the reporting of the spans. I would like to integrate the library in a StackStorm project (the underlying stack includes the usage of gunicorn and eventlet). My approach is the instantiation of a jaeger client (as a member of a custom singleton class) and the reporting of the spans is done inside a decorator used in the functions I would like to measure. The problem is that I must use time.sleep(2) in order for the spans to reach the collector. Is there a different approach to overcome this?

```
def tracing_decorator(func):
    @wraps(func)
    def wrapper_func(self, *args, **kwargs):
        name = func.__name__
        class_name = self.__class__.__name__
        with Tracer.get_instance().tracer_client.start_span(name) as span:
            span.log_kv({'function': name, 'class': class_name})
            s = func(self, *args, **kwargs)
        time.sleep(2)
        return s
    return wrapper_func
```

yurishkuro:
There is no different approach, I think, we just need to fix this specific issue.

obataku:
[...]

Issue #246, opened 2019-03-27
Support flushing span queue at process exit


nziebart:
## Requirement - what kind of business use case are you trying to solve?

We would like to ensure all tracing data is flushed to the backend before exiting the Python process. We have many jobs that are short-lived python processes, and the entire process is traced. Since the very last call to `span.finish()` happens right before the process exits, that span does not make it to the backend, because it is still in the in-memory span queue.

## Problem - what in Jaeger blocks you from solving the requirement?

There is no supported way to flush the in-memory queue, or wait until it is flushed.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Add a `tracer.flush()` method that will ensure all currently enqueued spans are sent to the backend before returning.

Probably the simplest solution is to call `Reporter._submit` synchronously on the entire queue. Alternatively we could wait until the length of the queue is zero, but that does not necessarily mean that the background thread has finished sending all the data.

We currently implement this in some apps by simply calling `time.sleep(3)`

## Any open questions to address

An edge case here could be if some background threads continue to submit spans while the `flush` method is being called. I think we could make the contract of this function that it will only flush spans that have been enqueued prior to the invocation of the method.

We may also want to provide a timeout, in case the function blocks when trying to acquire the flush lock (implementation could be acquiring the lock with the timeout).

presize-florian:
Any updates on this?

vlad-ivanov-name:
For anyone stumbling upon this in 2021, opentelemetry library doesn't seem to have this problem. I was able to reproduce this example with no issues: https://opentelemetry-python.readthedocs.io/en/latest/getting-started.html#configure-exporters-to-emit-spans-elsewhere

PR #294, opened 2020-12-01
Ensure Tracer.close() flushes finished spans

obataku:
* assigning to Reporter._stopped in _flush() instead of close() avoids dropping those spans passed to report_span just _prior_ to Reporter.close() for which _report_span_from_ioloop will execute only *afterwards*; this way those scheduled callbacks will get a chance to run before _flush stops the reporter and those pending spans make it into the queue before the reporter shuts down.

Signed-off-by: obataku <19821199+obataku@users.noreply.github.com>

this PR stops `Reporter.close()` from dropping spans whose background `_report_span_from_ioloop` callbacks are still pending execution; see #50 & #246


obataku:
I broke out the `Reporter.close()` fix from my draft PR #293

yurishkuro:
We need to fix #295 to ensure that all tests run.

yurishkuro (approved):

yurishkuro:
Thanks, @obataku. Do you think this resolves #50?

obataku:
@yurishkuro it _should_ resolve #50 so long as you `yield`/`await` the resultant `Future` in the context of the same IO/event loop as the tracer and are only concerned with flushing all spans `finish()`-ed prior to `close()`, i.e.
```py
tracer = Config(...).initialize_tracer(io_loop=IOLoop.current())
with tracer.start_active_span('foo'):
    ...
await tracer.close()
```

Issue #295, opened 2020-12-02
Migrate CI from Travis to GitHub Actions


yurishkuro:
We're having issues with Travis across jaegertracing org and need to move to GH actions.

Similar ticket (with many PRs) in the backend repo: https://github.com/jaegertracing/jaeger/issues/2645

* [x] convert unit tests #296
* [x] convert crossdock test
* [ ] update badges in the README
* [x] change required checks to new GH jobs
* [x] remove travis config

PR #296, opened 2020-12-03
Add github action for CI tests

### LINKED ISSUE #123: Config has no input verification

Ashmita152:
Signed-off-by: Ashmita Bohara <ashmita.bohara152@gmail.com>

Part of https://github.com/jaegertracing/jaeger-client-python/issues/295

TravisCI used to set python version using virtualenv, so makefile's bootstrap had a check while GitHub Actions doesn't use virtualenv, so we had to remove that check.

/cc @yurishkuro

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -1,0 +1,64 @@
+name: CI Tests
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  ci-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        version:
+        - python: "2.7"
+          tornado: ">=4,<5"
+        - python: "2.7"
+          tornado: ">=5,<6"
+        - python: "3.5"
+          tornado: ">=4,<5"
+        - python: "3.5"
+          tornado: ">=5,<6"
+        - python: "3.5"
+          tornado: ">=6"
+        - python: "3.6"
+          tornado: ">=4,<5"
+        - python: "3.6"
+          tornado: ">=5,<6"
+        - python: "3.6"
+          tornado: ">=6"
+        - python: "3.7"
+          tornado: ">=4,<5"
+        - python: "3.7"
+          tornado: ">=5,<6"
+        - python: "3.7"
+          tornado: ">=6"
+    name: ci-tests (py ${{ matrix.version.python }} tornado ${{ matrix.version.tornado }})
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+
+    - uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.version.python }}
+
+    - name: Export TORNADO variable
+      run: |
+        echo "TORNADO=${{ matrix.version.tornado }}" >> $GITHUB_ENV
+
+    - name: Install pycurl dependencies
+      run: sudo apt-get install libgnutls28-dev libcurl4-openssl-dev libssl-dev
+
+    - name: Install dependencies
+      run: make bootstrap
+
+    - name: Run tests
+      run: make test_ci
+
+    - name: Upload coverage to codecov
+      uses: codecov/codecov-action@v1
+      with:
+        verbose: true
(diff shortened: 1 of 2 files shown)


Ashmita152:
Sample Run: https://github.com/Ashmita152/jaeger-client-python/pull/1/checks

Makefile
```diff
@@ -18,7 +18,6 @@ PY_PATH = $(PYTHONPATH):$(PWD)
 
 .PHONY: bootstrap
 bootstrap:
-	[ "$$VIRTUAL_ENV" != "" ]
```
yurishkuro, line 21: Could we keep it and just set to some fake value in the actions? Having this check is very useful in local development, you don't want to accidentally install stuff globally (but it's fine in CI).

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+      run: sudo apt-get install libgnutls28-dev libcurl4-openssl-dev libssl-dev
+
+    - name: Install dependencies
+      run: make bootstrap
+
+    - name: Run tests
+      run: make test_ci
+
+    - name: Upload coverage to codecov
+      uses: codecov/codecov-action@v1
```
yurishkuro, line 61: Does this action support accumulation of coverage from multiple runs? I think there are some code branches that are sensitive to the versions matrix, and we compute coverage in every run anyway, it would make sense to aggregate.

Ashmita152: From https://docs.codecov.io/docs/merging-reports 
> Automatic report merging for all CI and languages into a single report.

Seems like codecov already does that.

yurishkuro: Yes, but the example in that link only calls codecov once, to submit multiple results in one bacth. In our case each job will call it, and I am not sure how codecov will know that those are all for the same PR, unless the action itself is smart enough to recognize that.

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+
+    - name: Export TORNADO variable
+      run: |
+        echo "TORNADO=${{ matrix.version.tornado }}" >> $GITHUB_ENV
+
+    - name: Install pycurl dependencies
+      run: sudo apt-get install libgnutls28-dev libcurl4-openssl-dev libssl-dev
+
+    - name: Install dependencies
+      run: make bootstrap
```
yurishkuro, line 57: do you need to echo VIRTUALENV to GITHUB_ENV first?

Ashmita152: Since we are setting VIRTUALENV at job level in #L11-12  it will automatically percolate down to all the steps.

.github/workflows/ci-tests.yml
Ashmita152: Not sure how the merged report should look like, but I can see the 11 uploads from the 11 CI jobs - https://codecov.io/gh/Ashmita152/jaeger-client-python/commit/092d2fe6a6f625924178bc503bbb2beac7aa0b5b/build 

And looking at the repo for pyra/cryptography (which they have mentioned in the above doc) https://codecov.io/gh/pyca/cryptography/commit/7d60786a72ed09fc53b5f7bd5e2c89166dccbbfa/build

Both look similar to me. 

From above doc: 

> Codecov does not override report data for multiple uploads. We always merge the data. Simply upload all three reports at once, or separately.

This feels like we don't need to do anything special on our side and just upload.

yurishkuro (approved):
Thanks!

yurishkuro:
> Part of #295

What else is remaining (besides badges in the README)?

yurishkuro:
Ah, I see, crossdock is remaining. I added a checklist to the ticket.

btw, the actions on master seem to have failed. I will try restarting

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,65 @@
+name: CI Tests
```
yurishkuro, line 1: let's rename this consistently with the main Jaeger repo: Unit Tests / unit-tests / unit-tests.yml

Ashmita152: Sure. I will do as part of crossdock PR. Thank you.

PR #298, opened 2020-12-06
Cleanup and feedbacks changes

### LINKED ISSUE #123: Config has no input verification

Ashmita152:
Signed-off-by: Ashmita Bohara <ashmita.bohara152@gmail.com>

Part of https://github.com/jaegertracing/jaeger-client-python/issues/295

- Removes `gh-3` tags
- Removes travis.yml and install-crossdock-deps.sh script which was used to setup a particular version of docker-compose
- Prevents double build of crossdock image by reusing existing tag.


scripts/build-crossdock.sh
```diff
@@ -21,10 +21,7 @@ else
   exit 0
 fi
 
-docker build -f crossdock/Dockerfile \
-	--build-arg tornado=$TORNADO \
-	--tag $REPO:$COMMIT .
-
+# This image is already built in `make crossdock` step
+docker tag crossdock_python:latest $REPO:$COMMIT
```
yurishkuro, line 25: Q: where is the name `crossdock_python:latest` coming from? It seems some implicit default behavior of docker-compose. If so, is it possible to give it a name explicitly instead (see next comment for why)?

scripts/build-crossdock.sh
```diff
@@ -21,10 +21,7 @@ else
...
 docker tag $REPO:$COMMIT $REPO:$TAG
-docker tag $REPO:$COMMIT $REPO:gh-$GITHUB_RUN_NUMBER
 docker push $REPO
```
yurishkuro, line 27: Q: how would docker know that we're pushing `crossdock_python:latest`? Seems like a name mismatch.

scripts/build-crossdock.sh
Ashmita152: Right it seems some implicit behaviour (crossdock is the directory name where docker-compose file exists and python is the service name within that docker_compose file). There should be some way to name it explicitly. Are you thinking about situation when changing something in docker-compose cause this script to fail since things are implicit.

scripts/build-crossdock.sh
Ashmita152: Docker will not push crossdock_python. It will only push all the tags for `$REPO` image i.e. `jaegertracing/xdock-py:<commit>` and `jaegertracing/xdock-py:<tag>`

scripts/build-crossdock.sh
yurishkuro: Yes, but I am not sure it would work even now, because when pushing you're referring to a different name altogether.

If docker-compose does not allow to override the name, we could just add an explicit build step to `make crossdock` and refer to that image from the docker-compose.yml

Ashmita152: From: https://docs.docker.com/compose/reference/build/

> If the Compose file specifies an image name, the image is tagged with that name

Hence, adding the image name in docker-compose should build image with that name.

Ashmita152:
Minor cleanup for Dockerfile:

- COPY can accept multiple source files (ref: https://docs.docker.com/engine/reference/builder/#copy)
- WORKDIR should create directory if it doesn't exist (ref: https://docs.docker.com/engine/reference/builder/#workdir)
- pip can be passed multiple -r options (ref: https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-r)
 
Less number of intermediate layers means faster build of docker images.

yurishkuro (approved):
Awesome!

PR #301, opened 2021-01-18
fix B3Codec extract error for b3 header

### LINKED ISSUE #123: Config has no input verification

MuchContact:
-

-

### DIFF
--- a/jaeger_client/codecs.py
+++ b/jaeger_client/codecs.py
@@ -296,7 +296,7 @@ class B3Codec(Codec):
             carrier[self.sampled_header] = '1'
 
     def extract(self, carrier):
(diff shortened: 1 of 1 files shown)


jaeger_client/codecs.py
```diff
@@ -296,7 +296,7 @@ def inject(self, span_context, carrier):
             carrier[self.sampled_header] = '1'
 
     def extract(self, carrier):
-        if not isinstance(carrier, dict):
+        if not hasattr(carrier, 'items'):
```
yurishkuro, line 299: it doesn't seem like it'd work in py2. maybe `six` has an equivalent test function

PR #304, opened 2021-04-26
Implement Binary codec

### LINKED ISSUE #224: tracer.inject with Format.BINARY is always empty

Injecting context data via tracer.inject in binary format. Trying to make the python cassandra driver compatible with [cassandra zipkin tracer](https://github.com/thelastpickle/cassandra-zipkin-tracing) by propagating context to Cassandra too.

```python
carrier = bytearray()  # always empty
tracer.inject(span, Format.BINARY, carrier)
```

I tried adapting the files in [Solution 3 of the tutorial](https://github.com/yurishkuro/opentracing-tutorial/tree/master/python/lesson03/solution) as well but binary is also empty.

I'm not really sure what causes the problem because TEXT_MAP or HTTP_HEARDS work just fine and return a `uber-trace-id`.

## Any open questions to address

opentracing==2.0.0
opentracing-instrumentation==2.4.3
Using this jaeger: `-e git+https://github.com/yurishkuro/jaeger-client-python.git@d631596415cb549a4e8314cc25d6c53fab710528#egg=jaeger_client`

cedy:
Resolves #224 

Implements extract and inject method of the Binary codec to be able to
inject/extract span context to/from bytearray.

This commit implements binary context injection/extraction the same way as in jaeger-client for go. Inject method dumps span context into bytearray carrier and packs it in big endian order as follow:
8 bytes uint64 for first 8 bytes of trace id, 8 bytes uint64 for last 8 bytes of trace id, 8 bytes uint64 for span id, 8 bytes uint64 for parent id, 1 byte uint8 for flags, 4 bytes unit32 for number of baggage items.
If baggage exists, its packs it as follows:
4 bytes unit32 for number of bytes in key-name, key-name bytes, 4 bytes unit32 for number of bytes in value, value-bytes.
Extract method follows the same pattern.

Binary codec is compatible with golang jaeger-client implementation.

### DIFF
--- a/tests/test_codecs.py
+++ b/tests/test_codecs.py
@@ -444,17 +444,14 @@ class TestCodecs(unittest.TestCase):
                  'baggage_count': 0,
                  'baggage': {}},
             b'a]\x85\xe0\xe0\x06\xd5[6k\x9d\x86\xaa\xbc\\\x8f#c\x06\x80jV\xdf\x826k\x9d\x86\xaa\xbc'
-            b'\\\x8f\x01\x00\x00\x00\x03\x00\x00\x00\x07key_one\x00\x00\x00\tvalue_one\x00\x00\x00'
-            b'\tkey_three\x00\x00\x00\x0bvalue_three\x00\x00\x00\x07key_two\x00\x00\x00\tvalue_two':
+            b'\\\x8f\x01\x00\x00\x00\x01\x00\x00\x00\x07key_one\x00\x00\x00\tvalue_one':
                 {'trace_id_high': 7015910995390813531,
                  'trace_id_low': 3921401102271798415,
                  'span_id': 2549888962631491458,
                  'parent_id': 3921401102271798415,
                  'flags': 1,
-                 'baggage_count': 3,
-                 'baggage': {'key_one': 'value_one',
-                             'key_two': 'value_two',
-                             'key_three': 'value_three'}
+                 'baggage_count': 1,
+                 'baggage': {'key_one': 'value_one'}
                  },
         }
 
--- a/tests/test_codecs.py
+++ b/tests/test_codecs.py
@@ -384,6 +384,31 @@ class TestCodecs(unittest.TestCase):
         with self.assertRaises(InvalidCarrierException):
             codec.extract({})
 
+        tracer = Tracer(
+            service_name='test',
+            reporter=InMemoryReporter(),
+            sampler=ConstSampler(True),
+        )
+        baggage = {'baggage_1': u'data',
+                   u'baggage_2': 'foobar',
+                   'baggage_3': '\x00\x01\x09\xff',
+                   u'baggage_4': u'\U0001F47E'}
+
+        span_context = SpanContext(trace_id=260817200211625699950706086749966912306, span_id=567890,
+                                   parent_id=1234567890, flags=1,
+                                   baggage=baggage)
+
+        carrier = bytearray()
+        tracer.inject(span_context, Format.BINARY, carrier)
+        assert len(carrier) != 0
+
+        extracted_span_context = tracer.extract(Format.BINARY, carrier)
+        assert extracted_span_context.trace_id == span_context.trace_id
+        assert extracted_span_context.span_id == span_context.span_id
+        assert extracted_span_context.parent_id == span_context.parent_id
+        assert extracted_span_context.flags == span_context.flags
+        assert extracted_span_context.baggage == span_context.baggage
+
 
 def test_default_baggage_without_trace_id(tracer):
     _test_baggage_without_trace_id(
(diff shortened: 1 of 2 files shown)


tests/test_codecs.py
```diff
@@ -384,6 +384,31 @@ def test_binary_codec(self):
...
+        carrier = bytearray()
+        tracer.inject(span_context, Format.BINARY, carrier)
+        assert len(carrier) != 0
+
+        extracted_span_context = tracer.extract(Format.BINARY, carrier)
+        assert extracted_span_context.trace_id == span_context.trace_id
+        assert extracted_span_context.span_id == span_context.span_id
+        assert extracted_span_context.parent_id == span_context.parent_id
+        assert extracted_span_context.flags == span_context.flags
+        assert extracted_span_context.baggage == span_context.baggage
```
yurishkuro, line 410: ideally the crossdock integration tests would have a mode to test binary propagation, but since we don't have that, one cheap substitution that I would suggest is serializing a few random contexts using Go client and hardcoding the binary representations into this test to compare against. It's not great for maintenance, but will at least somewhat test wire compatibility, whereas this current test only tests that the transform is not lossy, but that the transform is correct / compatible with Go.

cedy: Thank you for quick review. Added go compatibility test

tests/test_codecs.py
```diff
@@ -384,6 +384,95 @@ def test_binary_codec(self):
...
+            trace_id_high = (span_context.trace_id >> 64) & max_int64
+            trace_id_low = span_context.trace_id & max_int64
+
+            assert trace_id_high == expected["trace_id_high"]
+            assert trace_id_low == expected["trace_id_low"]
+            assert span_context.span_id == expected["span_id"]
+            assert span_context.parent_id == expected["parent_id"]
+            assert span_context.flags == expected["flags"]
+            assert len(span_context.baggage) == expected["baggage_count"]
+            assert span_context.baggage == expected["baggage"]
```
yurishkuro, line 474: I would add a call to `tracere.inject` here and compare the result with `span_context_serialized`

cedy: Added that, thanks

yurishkuro:
please make sure `make lint` is successful

cedy:
> please make sure `make lint` is successful

Lint issues are now resolved

yurishkuro:
Weird, one of the tests failed, PTAL. I would've thought this change would be mostly language-version agnostic.

## Unit Tests / unit-tests (py 3.5 tornado >=4,<5) (pull_request) 
```
____ TestCodecs.test_binary_codec_extract_compatibility_with_golang_client _____
tests/test_codecs.py:479: in test_binary_codec_extract_compatibility_with_golang_client
    assert carrier == bytearray(span_context_serialized)
E   AssertionError: assert bytearray(b'a...0\tvalue_one') == bytearray(b'a]...0\tvalue_two')
E     At index 40 diff: 9 != 7
E     Use -v to get the full diff

```

cedy:
> Weird, one of the tests failed, PTAL. I would've thought this change would be mostly language-version agnostic.
> 
> ## Unit Tests / unit-tests (py 3.5 tornado >=4,<5) (pull_request)
> ```
> ____ TestCodecs.test_binary_codec_extract_compatibility_with_golang_client _____
> tests/test_codecs.py:479: in test_binary_codec_extract_compatibility_with_golang_client
>     assert carrier == bytearray(span_context_serialized)
> E   AssertionError: assert bytearray(b'a...0\tvalue_one') == bytearray(b'a]...0\tvalue_two')
> E     At index 40 diff: 9 != 7
> E     Use -v to get the full diff
> ```

The problem is that python has deterministic dictionary order only from 3.7 version [https://stackoverflow.com/a/14959001](https://stackoverflow.com/a/14959001) . So, if we have more than 1 item in baggage, we can't determine how they are going to be dumped into byte array, therefore byte arrays don't align.  I don't see how to find the way around it, but only to have 1 item in the baggage.

yurishkuro:
Fair enough. Another failure is with Python 2.7:
```
_________________________ TestCodecs.test_binary_codec _________________________
tests/test_codecs.py:405: in test_binary_codec
    extracted_span_context = tracer.extract(Format.BINARY, carrier)
jaeger_client/tracer.py:273: in extract
    return codec.extract(carrier)
jaeger_client/codecs.py:182: in extract
    key, value, bytes_read = self._unpack_baggage_item(baggage_data)
jaeger_client/codecs.py:206: in _unpack_baggage_item
    value, b_read = self._read_kv(baggage[bytes_read:])
jaeger_client/codecs.py:214: in _read_kv
    return b''.join(data_value).decode('utf-8'), bytes_read
/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/encodings/utf_8.py:16: in decode
    return codecs.utf_8_decode(input, errors, True)
E   UnicodeDecodeError: 'utf8' codec can't decode byte 0xff in position 3: invalid start byte
```

I am not opposed to dropping 2.7 from CI since it's past EOL since 2020-01-01, but we'd need a different PR to remove it first before merging this one.

yurishkuro:
#306

yurishkuro (approved):

PR #314, opened 2021-07-15
Specify minimal required python version

### LINKED ISSUE #311: python_requires should be specified

Installing jaeger-client in Python 2

jaeger-client>=4.5.0 is not support Python 2.x, but still could installed by pip on this Python because lack of `python_requires` in setup.py.

It cause the following problem:
```
.../.tox/py27/lib/python2.7/site-packages/jaeger_client/codecs.py:32: in <module>
    import urllib.parse
E   ImportError: No module named parse
```

`python_requires` should be specified in setup.py to prevent pip from installing new packages for Python 2.x

ethframe:
- Resolves #311

- Specify minimal required python version using `python_requires`, as suggested in linked issue. To completely resolve this issue, the current latest package (4.5.0) should be removed from PyPI

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,7 @@ setup(
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
     ],
+    python_requires='>=3.5',
     install_requires=[
         'threadloop>=1,<2',
         'thrift',


ethframe:
> current latest package (4.5.0) should be removed from PyPI

On second thought, removing package from PyPI is a bad idea, and it would be better to release a new fixed package with version 4.6.0, and then reupload 4.4.0 as 4.5.1.

condorcet:
I think it would be better to release post-version of existing package according to PEP 440 https://www.python.org/dev/peps/pep-0440/#post-releases The changes are corresponding semantic of post-release, because it's actually not a bug fix
So latest release will be 4.5.0.post1

yurishkuro (approved):

yurishkuro:
Why can't we just release a new version?

ethframe:
> 
> 
> Why can't we just release a new version?

Because pip on python 2.7 will still install the incompatible 4.5.0 if user doesn't pin specific version or there is no newer package that actually works on 2.7.

PR #322, opened 2021-08-28
Fix B3Codec's trace_id zero-padding for 32bit IDs

mizhexiaoxiao:
- Solve the problem of incorrect traceid injection when the traceid is 128bit and starts with 0

For example, if the traceid is 023f352eaefd8b887a06732f5312e2de, it will inject as 23f352eaefd8b887a06732f5312e2de, which will cause the trace to split into two because the traceid is different

Signed-off-by: mizhexiaoxiao 1157861072@qq.com

### DIFF
--- a/jaeger_client/codecs.py
+++ b/jaeger_client/codecs.py
@@ -336,7 +336,7 @@ class B3Codec(Codec):
 
     def __init__(self, generate_128bit_trace_id=False):
         self.generate_128bit_trace_id = generate_128bit_trace_id
-      
+
     def inject(self, span_context, carrier):
         if not isinstance(carrier, dict):
             raise InvalidCarrierException('carrier not a dictionary')
--- a/jaeger_client/codecs.py
+++ b/jaeger_client/codecs.py
@@ -334,10 +334,16 @@ class B3Codec(Codec):
     flags_header = 'X-B3-Flags'
     _flags_header_lc = flags_header.lower()
 
+    def __init__(self, generate_128bit_trace_id=False):
+        self.generate_128bit_trace_id = generate_128bit_trace_id
+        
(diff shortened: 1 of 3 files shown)


jaeger_client/codecs.py
```diff
@@ -334,10 +334,16 @@ class B3Codec(Codec):
...
 
+    def __init__(self, generate_128bit_trace_id=False):
+        self.generate_128bit_trace_id = generate_128bit_trace_id
+        
     def inject(self, span_context, carrier):
         if not isinstance(carrier, dict):
             raise InvalidCarrierException('carrier not a dictionary')
-        carrier[self.trace_header] = format(span_context.trace_id, 'x').zfill(16)
+        if self.generate_128bit_trace_id:
+            carrier[self.trace_header] = format(span_context.trace_id, 'x').zfill(32)
```
yurishkuro, line 344: could you add a unit test that demonstrates the issue you're fixing?

yurishkuro:
Also, please fix the DCO (see CONTRIBUTING.md)

mizhexiaoxiao:
Thanks for the review!
Unit tests have been added

yurishkuro:
The DCO check is still failing, please sign all commits (see CONTRIBUTING)

yurishkuro (approved):

mizhexiaoxiao:
Thank you for your patient reply, it should be ok now

yurishkuro:
You can run "make lint" locally before submitting

mizhexiaoxiao:
done :)

yurishkuro:
Thanks!

yurishkuro:
https://github.com/jaegertracing/jaeger-client-python/releases/tag/4.6.1

Issue #323, opened 2021-09-03
Manage specific Sampler with Jaeger Config


Etienne-Carriere:
## Requirement - what kind of business use case are you trying to solve?

I would like to use custom Sampler with Jaeger Config (builder)

## Problem - what in Jaeger blocks you from solving the requirement?

Currently in Jaeger Config, it is only possible to set a dict to sampler parameter and so not able to able custom Sampler

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Accept in Jaeger Config a Sampler already configured

yurishkuro:
released in 4.7.0

PR #324, opened 2021-09-03
In Jaeger Config, enable to add directly a sampler

Etienne-Carriere:
It can be useful when using a specific sampler that is not included in
this library (for example private sampler)

https://github.com/jaegertracing/jaeger-client-python/issues/323

Accept on Jaeger Config parameter sampler not only a dict but also a Sampler object

### DIFF
--- a/jaeger_client/config.py
+++ b/jaeger_client/config.py
@@ -220,7 +220,7 @@ class Config(object):
     @property
     def sampler(self):
         sampler_param = self.config.get('sampler', {})
-        if isinstance(sampler_param , Sampler):
+        if isinstance(sampler_param, Sampler):
             return sampler_param
         sampler_type = sampler_param.get('type', None)
         sampler_param = sampler_param.get('param', None)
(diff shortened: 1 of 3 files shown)


jaeger_client/config.py
```diff
@@ -219,9 +219,11 @@ def max_traceback_length(self):
 
     @property
     def sampler(self):
-        sampler_config = self.config.get('sampler', {})
-        sampler_type = sampler_config.get('type', None)
-        sampler_param = sampler_config.get('param', None)
+        sampler_param = self.config.get('sampler', {})
```
yurishkuro, line 222: this is overloading `sampler_param` variable. I would keep it as `sampler_config`

Etienne-Carriere: I choose to change from `sampler_config` to `sampler_param` as now it can be both a sampler or a config so I choose a more generic name  but if you want I can come back to `sampler_config`

yurishkuro: I'd be fine with `sampler_config` or `sampler` (prefer config as it's more common use case). But `param` is confusing since it's the inner field.

Etienne-Carriere: you are right . I come back to sampler_config .

yurishkuro (approved):

Issue #319, opened 2021-08-11
Add type hints


kasium:
## Requirement - what kind of business use case are you trying to solve?

As you might know many python3 only packages already have type hints present based on the typing module. Afterwards the types can be checked e.g. with mypy. It would be great if you could add type hints and mark the package as types (aka add a py.typed file)

## Problem - what in Jaeger blocks you from solving the requirement?

Missing type hints in the code base

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Add type hints 😄

kasium:
@yurishkuro I'm happy to give a contribution a try. However, I guess with the current support python versions it will get hard. I would recommend to drop at least python 3.5 support if not even 3.6. What do you think?

yurishkuro:
yes, it's fine to raise min supported version (in separate PR)

kasium:
So, should I just change the setup.py file or also other stuff?

yurishkuro:
Also GitHub actions

kasium:
@yurishkuro so I see the following needed types

- [X] reporter ( #333 )
- [ ] config ( #334 )
- [X] util ( #326 )
- [X] tracer ( #328 )
- [ ] span
- [ ] metircs
- [ ] throttler
- [ ] sampler

Can you please check if something is missing or is too much

yurishkuro:
yep, looks just right

kasium:
@yurishkuro I currently hit an issue with my latest types (local change)

```
ImportError while loading conftest '/root/git/jaeger-client-python/tests/conftest.py'.
tests/conftest.py:17: in <module>
    from jaeger_client import ConstSampler, Tracer
jaeger_client/__init__.py:18: in <module>
    from .tracer import Tracer  # noqa
jaeger_client/tracer.py:35: in <module>
    from .span import Span, SAMPLED_FLAG, DEBUG_FLAG
jaeger_client/span.py:26: in <module>
    from .tracer import Tracer
E   ImportError: cannot import name 'Tracer' from 'jaeger_client.tracer' (/root/git/jaeger-client-python/jaeger_client/tracer.py)
```

This happens because `tracer.py` imports `span.py`, but `span.py` needs `tracer.py` to type the constructor. So I would need the `if TYPE_CHECKING` construct which will decrease the coverage

yurishkuro:
Does mypy have any means of adding types without explicit imports?

If no other solution, we'd have to take a hit on the coverage.

[...]

PR #326, opened 2021-09-07
Add basic typing support

kasium:
* Add myyp and main config
* Add flake8-typing-imports to avoid issues with the typing module
* Add some types to utils.py as a PoC

Part of #319

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,6 +4,7 @@ max-line-length = 100
 exclude =
   jaeger_client/thrift_gen/*,
   crossdock/thrift_gen/*
+min_python_version = 3.7.0
 
 [zest.releaser]
 release = no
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -21,3 +22,12 @@ python_files =
...
     tests.py
+
+[mypy]
+python_version = 3.7
+warn_unused_configs = True
+warn_redundant_casts = True
+warn_unused_ignores = True
+
+[mypy-six.*,tornado.*,opentracing.*,thrift.*,threadloop.*,prometheus_client.*]
+ignore_missing_imports = True
```
yurishkuro, line 33: What does this do?

kasium: So, by default mypy complains about imports to other libraries which have no types/type stubs. With that, I just tell mypy, that it should ignore it.

yurishkuro (approved):

setup.cfg
kasium: I agree. I guess there is already a open issue there

PR #328, opened 2021-09-07
Add typing to tracer.py

kasium:
This change was done by using pyannotate and running the test_tracer tests.
When in doubt, the type is now Any.

Part of #319 

### DIFF
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -22,7 +22,7 @@ import random
 import sys
 import time
 import opentracing
-from typing import Any, Optional, List, Union
+from typing import Any, Dict, Optional, List, Union
 
 from opentracing import Format, UnsupportedFormatException
 from opentracing.ext import tags as ext_tags
@@ -32,7 +32,7 @@ from opentracing.tracer import Reference
 from tornado.concurrent import Future
 
 from . import constants
-from .codecs import TextCodec, ZipkinCodec, ZipkinSpanFormat, BinaryCodec
+from .codecs import TextCodec, ZipkinCodec, ZipkinSpanFormat, BinaryCodec, Codec
 from .span import Span, SAMPLED_FLAG, DEBUG_FLAG
 from .span_context import SpanContext
 from .metrics import Metrics, LegacyMetricsFactory, MetricsFactory
@@ -58,8 +58,8 @@ class Tracer(opentracing.Tracer):
         baggage_header_prefix: str = constants.BAGGAGE_HEADER_PREFIX,
         debug_id_header: str = constants.DEBUG_ID_HEADER_KEY,
         one_span_per_rpc: bool = False,
-        extra_codecs: Optional[dict] = None,
-        tags: Optional[dict] = None,
+        extra_codecs: Optional[Dict[str, Codec]] = None,
+        tags: Optional[Dict[str, Any]] = None,
         max_tag_value_length: int = constants.MAX_TAG_VALUE_LENGTH,
         max_traceback_length: int = constants.MAX_TRACEBACK_LENGTH,
         throttler: Optional[Any] = None,
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -22,7 +22,7 @@ import random
 import sys
 import time
 import opentracing
-from typing import Any, Dict, Optional, List, Union
+from typing import Any, Dict, Optional, List, Union, cast
 
 from opentracing import Format, UnsupportedFormatException
 from opentracing.ext import tags as ext_tags
@@ -179,8 +179,9 @@ class Tracer(opentracing.Tracer):
         if valid_references and (parent is None or not parent.has_trace):
             parent = valid_references[0].referenced_context
 
-        rpc_server = tags and \
-            tags.get(ext_tags.SPAN_KIND) == ext_tags.SPAN_KIND_RPC_SERVER
+        rpc_server = cast(
+            bool, tags and tags.get(ext_tags.SPAN_KIND) == ext_tags.SPAN_KIND_RPC_SERVER
+        )
 
         if parent is None or not parent.has_trace:
             trace_id = self._random_id(self.max_trace_id_bits)
@@ -221,7 +222,7 @@ class Tracer(opentracing.Tracer):
                     operation_name=operation_name,
                     tags=tags, start_time=start_time, references=valid_references)
 
-        self._emit_span_metrics(span=span, join=bool(rpc_server))
+        self._emit_span_metrics(span=span, join=rpc_server)
 
         return span
 
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -22,16 +22,22 @@ import random
 import sys
 import time
 import opentracing
+from typing import Any, Optional, List, Union
+
 from opentracing import Format, UnsupportedFormatException
 from opentracing.ext import tags as ext_tags
-from opentracing.scope_managers import ThreadLocalScopeManager
+from opentracing.scope_managers import ThreadLocalScopeManager, ScopeManager
+from opentracing.tracer import Reference
+
+from tornado.concurrent import Future
 
 from . import constants
 from .codecs import TextCodec, ZipkinCodec, ZipkinSpanFormat, BinaryCodec
 from .span import Span, SAMPLED_FLAG, DEBUG_FLAG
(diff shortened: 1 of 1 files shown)


kasium:
Please note, that for most types I checked the tests and the docu, but I'm no expert in your code. Feel free to check and to provide better/correct types.

jaeger_client/tracer.py
```diff
@@ -209,18 +220,18 @@ def start_span(self,
                     operation_name=operation_name,
                     tags=tags, start_time=start_time, references=valid_references)
 
-        self._emit_span_metrics(span=span, join=rpc_server)
+        self._emit_span_metrics(span=span, join=bool(rpc_server))
```
yurishkuro, line 223: perhaps move to var declaration instead?

jaeger_client/tracer.py
```diff
@@ -41,19 +47,24 @@ class Tracer(opentracing.Tracer):
...
-        one_span_per_rpc=False, extra_codecs=None,
-        tags=None,
-        max_tag_value_length=constants.MAX_TAG_VALUE_LENGTH,
-        max_traceback_length=constants.MAX_TRACEBACK_LENGTH,
-        throttler=None,
-        scope_manager=None,
-    ):
+        self,
+        service_name: str,
+        reporter: Any,
```
yurishkuro, line 52: Should be BaseReporter that I am adding now in #329

jaeger_client/tracer.py
```diff
@@ -41,19 +47,24 @@ class Tracer(opentracing.Tracer):
...
+        sampler: Sampler,
+        metrics: Optional[Any] = None,
```
yurishkuro, line 54:
```suggestion
        metrics: Optional[Metrics] = None,
```

jaeger_client/tracer.py
```diff
@@ -41,19 +47,24 @@ class Tracer(opentracing.Tracer):
...
+        metrics_factory: Optional[MetricsFactory] = None,
+        trace_id_header: str = constants.TRACE_ID_HEADER,
+        generate_128bit_trace_id: bool = False,
+        baggage_header_prefix: str = constants.BAGGAGE_HEADER_PREFIX,
+        debug_id_header: str = constants.DEBUG_ID_HEADER_KEY,
+        one_span_per_rpc: bool = False,
+        extra_codecs: Optional[dict] = None,
+        tags: Optional[dict] = None,
```
yurishkuro, line 62: probably could be more specific like `Optional[Dict[str, Any]]`?

jaeger_client/tracer.py
```diff
@@ -41,19 +47,24 @@ class Tracer(opentracing.Tracer):
...
+        reporter: Any,
+        sampler: Sampler,
+        metrics: Optional[Any] = None,
+        metrics_factory: Optional[MetricsFactory] = None,
+        trace_id_header: str = constants.TRACE_ID_HEADER,
+        generate_128bit_trace_id: bool = False,
+        baggage_header_prefix: str = constants.BAGGAGE_HEADER_PREFIX,
+        debug_id_header: str = constants.DEBUG_ID_HEADER_KEY,
+        one_span_per_rpc: bool = False,
+        extra_codecs: Optional[dict] = None,
```
yurishkuro, line 61: maybe `Optional[Dict[str, Codec]]`?

jaeger_client/tracer.py
```diff
@@ -41,19 +47,24 @@ class Tracer(opentracing.Tracer):
...
+        tags: Optional[dict] = None,
+        max_tag_value_length: int = constants.MAX_TAG_VALUE_LENGTH,
+        max_traceback_length: int = constants.MAX_TRACEBACK_LENGTH,
+        throttler: Optional[Any] = None,
```
yurishkuro, line 65: need to add Throttler class and have RemoteThrottler extend it

yurishkuro (commented):
few nits, otherwise looks great

jaeger_client/tracer.py
kasium: Can you please give me an example what you mean?

jaeger_client/tracer.py
kasium: So, should I wait till this is done?

jaeger_client/tracer.py
yurishkuro: L182, `rpc_server: bool = ...`

jaeger_client/tracer.py
yurishkuro: if you want to add it, go ahead, but I'm ok to merge as is.

kasium: I see this outside of this PR to be honest 😄

jaeger_client/tracer.py
kasium: Need to use cast, but else it works

jaeger_client/tracer.py
```diff
@@ -167,8 +179,9 @@ def start_span(self,
         if valid_references and (parent is None or not parent.has_trace):
             parent = valid_references[0].referenced_context
 
-        rpc_server = tags and \
-            tags.get(ext_tags.SPAN_KIND) == ext_tags.SPAN_KIND_RPC_SERVER
+        rpc_server = cast(
+            bool, tags and tags.get(ext_tags.SPAN_KIND) == ext_tags.SPAN_KIND_RPC_SERVER
+        )
```
yurishkuro, line 184: when I run `make lint` it works fine without this change. How are you testing typing?

kasium: Did you also remove "bool" from the other line. I see then
```
jaeger_client/tracer.py:223: error: Argument "join" to "_emit_span_metrics" of "Tracer" has incompatible type "Union[Dict[Any, Any], None, Any]"; expected "Optional[bool]"
```

yurishkuro:
```
$ make lint
flake8 jaeger_client crossdock tests
mypy jaeger_client
Success: no issues found in 36 source files
./scripts/check-license.sh

$ gd | cat
diff --git a/jaeger_client/tracer.py b/jaeger_client/tracer.py
index e4e8e00..e553f6e 100644
--- a/jaeger_client/tracer.py
+++ b/jaeger_client/tracer.py
@@ -22,6 +22,8 @@ import random
 import sys
 import time
 import opentracing
+from typing import Optional
+
 from opentracing import Format, UnsupportedFormatException
 from opentracing.ext import tags as ext_tags
 from opentracing.scope_managers import ThreadLocalScopeManager
@@ -282,7 +284,7 @@ class Tracer(opentracing.Tracer):
         self.sampler.close()
         return self.reporter.close()

-    def _emit_span_metrics(self, span, join=False):
+    def _emit_span_metrics(self, span: Span, join: Optional[bool] = False):
         if span.is_sampled():
             self.metrics.spans_started_sampled(1)
         else:
```

yurishkuro: I did `pip install mypy`
```
mypy-0.910
mypy_extensions-0.4.3
```

kasium: So, of course if you remove the type annotation of the argument, mypy will not complain 😄

yurishkuro: I do have the type declared in the function. I also tried with my original suggestion, still successful:
```
        rpc_server: bool = tags and \
            tags.get(ext_tags.SPAN_KIND) == ext_tags.SPAN_KIND_RPC_SERVER
```

kasium: This one should fail, because a cast is needed. Not sure, why it's successful on your side. On mine it fails as expected. 
Maybe a weird mypy cache issue or something else in your setup. Might also be related to the python version which you running.

However, making a cast is fine, so I see no risk here

yurishkuro: Yes, but the cast is ugly code. What errors are you getting with `rpc_server: bool =` ?

I'm running with Python 3.8.2

kasium:
```
jaeger_client/tracer.py:182: error: Incompatible types in assignment (expression has type "Union[Dict[Any, Any], None, Any]", variable has type "bool")
```

kasium: Ok, found a way to write it easier

yurishkuro (approved):

jaeger_client/tracer.py
yurishkuro: #330

PR #331, opened 2021-09-07
Clean-up warnings

yurishkuro:
```
tests/test_crossdock.py:54
  /home/runner/work/jaeger-client-python/jaeger-client-python/tests/test_crossdock.py:54: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    def tracer():

tests/test_reporter.py:100
  /home/runner/work/jaeger-client-python/jaeger-client-python/tests/test_reporter.py:100: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    def thread_loop(self):

```

### DIFF
--- a/tests/test_crossdock.py
+++ b/tests/test_crossdock.py
@@ -50,7 +50,7 @@ def app():
 
 
 # noinspection PyShadowingNames
-@pytest.yield_fixture
+@pytest.fixture
 def tracer():
     tracer = Tracer(
         service_name='test-tracer',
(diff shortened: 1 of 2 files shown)


kasium:
@yurishkuro if you like I can approve this change 😄 Already had the same idea

yurishkuro:
I forgot about it

PR #332, opened 2021-09-08
Add more types (tracer, utils, thrift, span_context, tudp_transport)

kasium:
Part of #319 

### DIFF
--- a/jaeger_client/TUDPTransport.py
+++ b/jaeger_client/TUDPTransport.py
@@ -14,6 +14,7 @@
 
 from __future__ import absolute_import
 import logging
+from typing import Optional
 
 from thrift.transport.TTransport import TTransportBase
 import socket
@@ -26,30 +27,31 @@ class TUDPTransport(TTransportBase, object):
     TUDPTransport implements just enough of the tornado transport interface
     to work for blindly sending UDP packets.
     """
-    def __init__(self, host, port, blocking=False):
+    def __init__(self, host: str, port: int, blocking: bool = False) -> None:
         self.transport_host = host
         self.transport_port = port
-        self.transport_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        if blocking:
-            blocking = 1
-        else:
-            blocking = 0
+        self.transport_sock: Optional[socket.socket] = (
+            socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        )
         self.transport_sock.setblocking(blocking)
 
-    def write(self, buf):
+    def write(self, buf: bytes) -> int:
         """Raw write to the UDP socket."""
+        if not self.transport_sock:
+            raise ValueError('Socket is already closed')
         return self.transport_sock.sendto(
             buf,
             (self.transport_host, self.transport_port)
         )
 
-    def isOpen(self):
+    def isOpen(self) -> bool:
         """
         isOpen for UDP is always true (there is no connection) as long
         as we have a sock
         """
         return self.transport_sock is not None
 
-    def close(self):
-        self.transport_sock.close()
-        self.transport_sock = None
+    def close(self) -> None:
+        if self.transport_sock:
+            self.transport_sock.close()
+            self.transport_sock = None
--- a/jaeger_client/TUDPTransport.py
+++ b/jaeger_client/TUDPTransport.py
@@ -14,7 +14,6 @@
 
 from __future__ import absolute_import
 import logging
-from typing import Optional
 
 from thrift.transport.TTransport import TTransportBase
 import socket
@@ -27,31 +26,30 @@ class TUDPTransport(TTransportBase, object):
     TUDPTransport implements just enough of the tornado transport interface
     to work for blindly sending UDP packets.
     """
-    def __init__(self, host: str, port: int, blocking: bool = False) -> None:
+    def __init__(self, host, port, blocking=False):
         self.transport_host = host
         self.transport_port = port
-        self.transport_sock: Optional[socket.socket] = (
-            socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        )
+        self.transport_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        if blocking:
+            blocking = 1
+        else:
+            blocking = 0
         self.transport_sock.setblocking(blocking)
 
-    def write(self, buf: bytes) -> int:
+    def write(self, buf):
         """Raw write to the UDP socket."""
-        if not self.transport_sock:
-            raise ValueError('Socket is already closed')
         return self.transport_sock.sendto(
             buf,
             (self.transport_host, self.transport_port)
         )
 
-    def isOpen(self) -> bool:
+    def isOpen(self):
         """
         isOpen for UDP is always true (there is no connection) as long
         as we have a sock
         """
         return self.transport_sock is not None
 
-    def close(self) -> None:
-        if self.transport_sock:
-            self.transport_sock.close()
-            self.transport_sock = None
+    def close(self):
+        self.transport_sock.close()
+        self.transport_sock = None
--- a/jaeger_client/span_context.py
+++ b/jaeger_client/span_context.py
@@ -15,7 +15,6 @@
 from __future__ import absolute_import
 
 import opentracing
-from typing import Optional
 
 
 class SpanContext(opentracing.SpanContext):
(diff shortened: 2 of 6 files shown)


jaeger_client/thrift.py
```diff
@@ -13,37 +13,48 @@
...
+from typing import Any, List, Tuple, Optional, TYPE_CHECKING
+from types import TracebackType
+
+from opentracing.tracer import ReferenceType, Reference
 
+from .constants import MAX_TRACEBACK_LENGTH
 import jaeger_client.thrift_gen.jaeger.ttypes as ttypes
 import jaeger_client.thrift_gen.sampling.SamplingManager as sampling_manager
 
+if TYPE_CHECKING:
```
kasium, line 25: fyi, this syntax is needed to avoid circular imports. If a type from this section is used it must be wrapped in quotes. Mypy will the dynamically resolve the type

jaeger_client/span_context.py
```diff
@@ -49,20 +57,20 @@ def with_baggage_item(self, key, value):
         )
 
     @property
-    def has_trace(self):
-        return self.trace_id and self.span_id and self.flags is not None
+    def has_trace(self) -> bool:
+        return self.flags is not None and self.trace_id and self.span_id  # type: ignore
```
kasium, line 61: I have still no clue why some boolean expressions are not resolved as bools. Let's ignore this for now

kasium:
How can I ignore a line on Codecov? The `if TYPE_CHECKING` lines are for mypy only

jaeger_client/span_context.py
yurishkuro: I don't think they are Boolean: 'and x' in a short circuit situation returns the actual x

jaeger_client/span_context.py
```diff
@@ -15,14 +15,22 @@
...
     __slots__ = ['trace_id', 'span_id', 'parent_id', 'flags',
                  '_baggage', '_debug_id']
 
     """Implements opentracing.SpanContext"""
-    def __init__(self, trace_id, span_id, parent_id, flags, baggage=None, debug_id=None):
+    def __init__(
+        self, trace_id: Optional[int],
+        span_id: Optional[int],
+        parent_id: Optional[int],
+        flags: Optional[int],
```
yurishkuro, line 30: Hm, is there any code that actually creates context with None for these fields? We treat 0 as invalid value, but otherwise IDs are ints.

jaeger_client/thrift.py
```diff
@@ -13,37 +13,48 @@
 # limitations under the License.
 
 import traceback
```
yurishkuro, line 15: I thought your objective was to supply types for public APIs. Do we need to touch these internal files?

jaeger_client/tracer.py
```diff
@@ -37,9 +37,11 @@
 from .span_context import SpanContext
 from .metrics import Metrics, LegacyMetricsFactory, MetricsFactory
 from .utils import local_ip
-from .sampler import Sampler
-from .reporter import BaseReporter
-from .throttler import Throttler
+
+if TYPE_CHECKING:
```
yurishkuro, line 41: What was wrong explicit imports?

jaeger_client/utils.py
```diff
@@ -27,12 +31,17 @@ class ErrorReporter(object):
     N.B. metrics will be deprecated in the future
     """
 
-    def __init__(self, metrics, logger=None, log_interval_minutes=15):
+    def __init__(
+        self,
+        metrics: Optional['Metrics'],
```
yurishkuro, line 36: Doesn't look like it's used, so Any is fine, and a comment "not used"

jaeger_client/TUDPTransport.py
```diff
@@ -26,30 +27,31 @@ class TUDPTransport(TTransportBase, object):
...
+    def __init__(self, host: str, port: int, blocking: bool = False) -> None:
         self.transport_host = host
         self.transport_port = port
-        self.transport_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        if blocking:
-            blocking = 1
-        else:
-            blocking = 0
+        self.transport_sock: Optional[socket.socket] = (
+            socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
```
yurishkuro, line 34: This can return None?

jaeger_client/thrift.py
kasium: Can you please tell me then, which files are considered "public"?

jaeger_client/tracer.py
kasium: It can lead to import cycles. But let me remove it

jaeger_client/TUDPTransport.py
kasium: Nope, but the variable can get None. If you call close, you set the variable to None

jaeger_client/span_context.py
kasium: Yes, they can get None: SpanContext.with_debug_id()

jaeger_client/thrift.py
yurishkuro: Strictly speaking, the only APIs from Jaeger that users need to use directly are the config.py module and the Tracer constructor, all other integrations with the SDK should be via OpenTracing API. However, Tracer constructor takes a lot of other internal types like Sampler / Reporter, so that technically increases the scope of how much typing is needed to help users.

kasium: So span_context and TUDPTransport can be considered internal?

jaeger_client/TUDPTransport.py
```diff
@@ -26,30 +27,31 @@ class TUDPTransport(TTransportBase, object):
...
     """
-    def __init__(self, host, port, blocking=False):
+    def __init__(self, host: str, port: int, blocking: bool = False) -> None:
         self.transport_host = host
         self.transport_port = port
-        self.transport_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        if blocking:
-            blocking = 1
-        else:
-            blocking = 0
```
yurishkuro, line 36: Is this safe to remove? Does transport_sock.setblocking accept booleans and ints?

I'd be more comfortable with keeping the existing logic, just folding it into an expression
```
self.transport_sock.setblocking(1 if blocking else 0)
```

jaeger_client/span_context.py
kasium: Of course, thanks 😄

jaeger_client/TUDPTransport.py
kasium: So according to the docs, it never accepted ints: https://docs.python.org/3/library/socket.html#socket.socket.setblocking

jaeger_client/span_context.py
yurishkuro: that function is deprecated, we can change it to initialize with 0s instead of None's if that's the only reason.

jaeger_client/span_context.py
```diff
@@ -15,14 +15,22 @@
...
+        baggage: Optional[dict] = None,
```
yurishkuro, line 31:
```suggestion
        baggage: Optional[Dict[str, str]] = None,
```

jaeger_client/tracer.py
```diff
@@ -51,8 +51,8 @@ class Tracer(opentracing.Tracer):
     def __init__(
         self,
         service_name: str,
-        reporter: BaseReporter,
-        sampler: Sampler,
+        reporter: 'BaseReporter',
+        sampler: 'Sampler',
```
yurishkuro, line 55: quotes not necessary with explicit imports, right?

jaeger_client/span_context.py
kasium: I reset the files anyway. Can be done in a separate commit (you might want to create an issue?)

jaeger_client/tracer.py
kasium: right

jaeger_client/tracer.py
```diff
@@ -37,9 +37,9 @@
 from .span_context import SpanContext
 from .metrics import Metrics, LegacyMetricsFactory, MetricsFactory
 from .utils import local_ip
-from .sampler import Sampler
```
yurishkuro, line 40: it doesn't look like anything is changing in this file at all

kasium:
Let me dump this PR and open a new, clean one

PR #333, opened 2021-09-08
Add types to reporter

kasium:

Part of #319

### DIFF
--- a/=5
+++ b/=5
@@ -1,1 +1,0 @@
-Requirement already satisfied: tornado in ./venv/lib64/python3.7/site-packages (5.1.1)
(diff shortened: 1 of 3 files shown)


kasium:
So hopefully this PR is better. I only added hints for public API methods/classes

yurishkuro (approved):

kasium:
@yurishkuro not sure what to do with the failing coverage. I didn't change the logic, but I guess the number of lines make there a difference

=5
```diff
@@ -0,0 +1 @@
+Requirement already satisfied: tornado in ./venv/lib64/python3.7/site-packages (5.1.1)
```
yurishkuro, line 1: ?

kasium: No clue where this somes from

yurishkuro (approved):

PR #334, opened 2021-09-08
Add types to config

kasium:

Part of #319

### DIFF
--- a/jaeger_client/config.py
+++ b/jaeger_client/config.py
@@ -152,8 +152,9 @@ class Config(object):
         return self._service_name
 
     @property
-    def metrics(self) -> Optional[Metrics]:
-        return self._metrics  # type:ignore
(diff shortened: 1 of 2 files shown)


jaeger_client/config.py
```diff
@@ -138,42 +148,42 @@ def _validate_config(self, config):
...
     @property
-    def service_name(self):
+    def service_name(self) -> str:
         return self._service_name
 
     @property
-    def metrics(self):
-        return self._metrics
+    def metrics(self) -> Optional[Metrics]:
+        return self._metrics  # type:ignore
```
yurishkuro, line 156: I don't think there is such thing as `self._metrics`, let's return None and add a comment "Deprecated"

kasium: done

yurishkuro (approved):

PR #335, opened 2021-09-08
Remove __future__ imports

kasium:
These imports are just not needed anymore, since we run on python 3.7+.
See here https://docs.python.org/3/library/__future__.html


yurishkuro (approved):

kasium:
Not sure what to do with the coverage...

yurishkuro:
looks like a flaky test somewhere which changes the coverage randomly

PR #339, opened 2021-09-12
Add ipv6 support

### LINKED ISSUE #337: IPv6 support
Currently jaeger_client is not flexible enough - [it is not possible](https://github.com/jaegertracing/jaeger-client-python/blob/master/jaeger_client/TUDPTransport.py#L31) to specify ipv6 socket.

Could you please add such support? I could do PR by myself, but i am not sure what is the best way to implement it. 

Of course, it would be much more flexible to pass socket object, that could be configured by user. But for current API perhaps it would be better to add one more parameter to the config:

```
config = Config(
    'local_agent': {
        'reporting_host': 'example.com',  # standard option
        'reporting_host_ipv6': 'example.com', # instead?
        'reporting_host_use_ipv6': true,  # one more way?
        'reporting_port': 6831,
    },
    ...
)
```

alvassin:
Added ability to report spans to ipv6 only hosts using:
- config `local_agent.reporting_host_use_ipv6` option
- `JAEGER_AGENT_HOST_USE_IPV6` environment variable.
- if none of above options are not specified, jaeger client checks if host behind `reporting_host` is ipv6 only.

Resolves #337

### DIFF
--- a/jaeger_client/TUDPTransport.py
+++ b/jaeger_client/TUDPTransport.py
@@ -17,6 +17,7 @@ import logging
 from thrift.transport.TTransport import TTransportBase
 import socket
 
+
 logger = logging.getLogger('jaeger_tracing')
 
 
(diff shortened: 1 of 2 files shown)


jaeger_client/utils.py
```diff
@@ -120,3 +120,16 @@ def interface_ip(interface: bytes) -> Optional[str]:
     # Explanation:
     # http://stackoverflow.com/questions/11735821/python-get-localhost-ip
     # http://stackoverflow.com/questions/24196932/how-can-i-get-the-ip-address-of-eth0-in-python
+
+
+def check_if_address_has_only_ipv6(address: str, port: Optional[int] = None) -> bool:
+    has_ipv6 = False
+    try:
+        for family, *_ in socket.getaddrinfo(address, port, type=socket.SOCK_DGRAM):
```
yurishkuro, line 128: this is not

jaeger_client/config.py
```diff
@@ -303,6 +303,25 @@ def local_agent_reporting_host(self) -> str:
         else:
             return DEFAULT_REPORTING_HOST
 
+    @property
+    def local_agent_reporting_host_use_ipv6(self) -> bool:
```
yurishkuro, line 307: Since you're automatically determine if the address is IPv6-only, what is the value proposition of this new property? To put it differently, what is the downside of not exposing it and always handling the v4/v6 automatically?

alvassin: Resonable. Added `_create_socket` method to TUDPTransport class, that can be overriden, removed config options.

Also, socket.set_blocking() requires boolean, so i removed the following block (perhaps there is a reason to keep it):
```
if blocking:
    blocking = 1
else:
    blocking = 0
```

alvassin: Added tests

